Issue #78, opened 2018-11-11
Enhancement: Filtering selecting rows


balibou:
example on: https://mbrn.github.io/material-table/#/docz-examples-example-selection

1. Select a row
2. Try to display only selected rows

When selecting a row with a checkbox, I would like to filter only selected rows

mbrn:
Good idea. I will work on it.

mbrn:
@balibou, 

Why you don't fork repo and try something like that?

balibou:
@mbrn I will have a look 😃

luizbaldi:
Hi @balibou, feel free to ask here if you have any doubts or suggestions 😄

balibou:
Hi @luizbaldi, I'm already working on a branch 😄Just a bit tricky because the checkbox column is not a proper column, but I should figure a good solution ! Will pull request it soon

balibou:
@luizbaldi @mbrn Can I get the access rights for pushing my PR ? 😃

mbrn:
Thanks for your contribution @balibou.

Did you fork project? You can send your changes with this order:

1. Fork the project
2. Clone the your own repo to the local.
3. Change what ever you want
4. Commit and push your changes to your repo.
5. Create a pull request. 
6. I will be very happy to accept it :)

balibou:
@mbrn ah ok actually I didn't fork it. Will do that then

mbrn:
This is in v1.3.0

PR #81, opened 2018-11-17
feat(filter): add checked checkbox filter

balibou:
## Related Issue
#78

## Description
When selecting a row with a checkbox, I would like to filter only selected rows

* m-table-body.js
* m-table-filter-row.js
* material-table.js
* package.json (@material-ui/core in dependencies missing)
* example-options.mdx (for updating demo)

## Additional Notes
Awesome package !


mbrn:
After i check everything  i will merge this. Thank you @balibou

balibou:
@mbrn hold on I just spotted something I forgot. I will fix it now

balibou:
@mbrn done 😄

mbrn:
Thanks @balibou , I hope that i can merge today

balibou:
@mbrn if you remove material core in package.json there will be an error when doing a yarn

mbrn:
> @mbrn if you remove material core in package.json there will be an error when doing a yarn

Yes but it is peedDependent package. So we wait user already have material-ui/core in their project

balibou:
> I added yarn.lock to .gitignore

you should let it in the repo (best practice), to be sure that everyone that work on the package have the same versions of the dependencies locally

mbrn:
> To me it should work like this:
> 
> Checked: Show only checked rows
> Unchecked: Show all rows (because the filter is not activated)
> Unstable: No state

Yes.  but our current boolean filter works as I mention about.

mbrn:
> > I added yarn.lock to .gitignore
> 
> you should let it in the repo (best practice), to be sure that everyone that work on the package have the same versions of the dependencies locally

I didnt know it is the best practice. I will look at this and add it again.

balibou:
> I didnt know it is the best practice. I will look at this and add it again.

https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

balibou:
> Yes. but our current boolean filter works as I mention about.

Ok maybe we should change it

mbrn:
Thanks @balibou. I will add yarn.lock again.

akshaycj47:
@balibou @mbrn is it still possible to do this? 
It looks the filter checkbox is replaced by select all checkbox?

PR #91, opened 2018-12-01
This closes #89, example-emptydatasource.mdx can be helpful about usi…

### LINKED ISSUE #89: Custom Text when there is no data in the table.
When there are no rows,I want to display a message like, "There is no data in the table " inside the table. Could you please check on that.

Thanks.

yildirims:
This closes #89, example-emptydatasource.mdx can be helpful about using showEmptyDataSourceMessage and emptyDataSourceMessage props to show message when there is no data to display.

### DIFF
--- a/docz/examples/example-emptydatasource.mdx
+++ b/docz/examples/example-emptydatasource.mdx
@@ -1,0 +1,35 @@
+---
+name: Empty DataSource Message Example
+order: 7
+menu: Examples
+---
+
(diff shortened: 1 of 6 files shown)


mbrn:
Hi @yildirims,

Thanks for your contribution. But before merging your PR can you make visual improvements on your code. 
1. Vertical and horizantal center
2. Vertical stability that means table should be same heigth either has data or not

PR #95, opened 2018-12-03
This closes #84

### LINKED ISSUE #84: AlphaNumberic Sort
Example: 
Sorting works fine on the following values -  112300, 34500, 54300

But the below example does a string sort instead of an alphanumeric search:

$112,300
$34,500
$54,300

Expected O/P : 
$34,500
$54,300
$112,300

Actual O/P:
$112,300
$34,500
$54,300

yildirims:
## Related Issue
Relate the Github issue with this PR using `#84`

### DIFF
--- a/docz/examples/example-currency.mdx
+++ b/docz/examples/example-currency.mdx
@@ -1,0 +1,53 @@
+---
+name: Currency Example
+order: 7
+menu: Examples
+---
+
+import { Playground } from 'docz'
+import MaterialTable from '../../src/material-table'
+
+# Currency Example
+
+<Playground>
+  <MaterialTable
+    columns={[
+      {title: 'Name', field: 'name'},
+      {title: 'Surname', field: 'surname'},
+      {title: 'Expenses', field: 'expenses', type: 'currency'},    
+    ]}
+    data={[
+      {name: 'Mehmet', surname: 'Baran', expenses:1500.50 },
+      {name: 'Selahattin', surname: 'Yıldırım', expenses:1200 },
+      {name: 'Silvia Joyce', surname: 'Balraj', expenses:1700 },
+    ]}
+    title="Currency"
+    options={{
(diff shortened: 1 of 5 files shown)


docz/examples/example-currency.mdx
```diff
@@ -0,0 +1,53 @@
+---
+name: Currency Example
```
mbrn: Can you change this file as column types example. We should make group the similar examples.

index.d.ts
```diff
@@ -26,7 +26,8 @@ export interface Column {
   lookup?: object;
   render?: (data:any) => any;
   title: string;
-  type?: 'boolean' | 'numeric' | 'date' | 'datetime' | 'time';
+  type?: 'boolean' | 'numeric' | 'date' | 'datetime' | 'time' | 'currency';
+  currencySetting?:object;
```
mbrn: Instead of object you should write something like {x: number, y: string.........} to make currencySetting pattern more clear

src/material-table.js
```diff
@@ -363,7 +363,8 @@ MaterialTable.propTypes = {
     lookup: PropTypes.object,
     render: PropTypes.func,
     title: PropTypes.string.isRequired,
-    type: PropTypes.oneOf(['boolean', 'numeric', 'date', 'datetime', 'time'])
+    type: PropTypes.oneOf(['boolean', 'numeric', 'date', 'datetime', 'time', 'currency']),
+    currencySetting:PropTypes.object
```
mbrn: Same with typescript review

index.d.ts
yildirims: Ok I'll refactor my code as you mentioned above

docz/examples/example-currency.mdx
yildirims: Ok.

Issue #98, opened 2018-12-04
Should be a option to filter on hidden fields


netbrain:
The following line:
https://github.com/mbrn/material-table/blob/master/src/material-table.js#L150

There should be a option to filter on hidden fields. Now the search functionality excludes fields that are hidden. But I would like ti include them in search, but not display them.

Should be simple enough to add a flag for this.

mbrn:
Hi @netbrain. 

We can add this option easily. We will do it. But why you dont contribute project with this :)

PR #100, opened 2018-12-05
Added a "searchable" property ❤️

netbrain:
## Related Issue
challenge accepted! #98  

## Description
Added a searchable property  which allows for overriding the default behavior of excluding hidden fields from search

Column definition and apply search function


netbrain:
@mbrn ready for review :+1:

Issue #82, opened 2018-11-23
Checked rows disappearing 


Monteyne:
Hello,
checked rows keep disappearing after receiving next props, 
i have looked into source, seems like new data from componentWillReceiveProps don't have "checked" property.

mbrn:
Hello @Monteyne,

You are right. This is a bug. I will resolve this asap.

balibou:
Hello @Monteyne @mbrn,
Is it possible to get a workflow please ?
Thanks !

mbrn:
1. Render table with table and onSelectionChange event. 
2. select some rows and this fires onSelectionChange event. In onSelectionChange event do something with setState on your component that uses table. So the component will rerender. Then checked rows will be disappeared.

balibou:
still working on it but didn't find a good solution

mbrn:
This problem resolved with #132. Thanks to @balibou

Issue #96, opened 2018-12-03
Enhancement: Init already checked row


balibou:
**Is your feature request related to a problem? Please describe.**
I would like to be able to load data with already checked rows

**Describe the solution you'd like**
Implement it 😄 (can try the ticket)

mbrn:
You can send already init checked row with adding tableData:{checked: true} to your row-data. 

```javascript
 data: [
      { name: 'Mehmet', surname: 'Baran', birthYear: 1987, birthCity: 63, tableData: { checked: true } },
      { name: 'Mehmet', surname: 'Baran', birthYear: 1987, birthCity: 63 },
      { name: 'Mehmet', surname: 'Baran', birthYear: 1987, birthCity: 63 },
    ]
```

sshanzel:
Is it working now? Cause i am implementing the tableData with checked property but it is not being checked.

PR #101, opened 2018-12-05
feat(check): possibility to load already checked rows

balibou:
## Related Issue
#96 

## Description
be able to load data with already checked rows

* example-selection.mdx
* material-table.js


balibou:
😉

mbrn:
But i have to revert this PR. Because firstly we should resolve #82 . And i think that when we resolve #82 this checked property should be implemented with another type. Such as defaultCheckedRows list.

balibou:
ok I will have a look on #82

Issue #70, opened 2018-11-04
Customize toolbar rendering


tadeuqc:
**Describe the solution you'd like**
rendering a custom toolbar would allow to render controls like the title and the search textbox in different orders, or even add button like add/edit Records in a modal.

**Describe alternatives you've considered**
It could be done creating a custom component and pointing it on a prop. So we could incorporate the behaviors with a totally custom UI

mbrn:
Hi @tadeuqc,

This is a good idea. I can add `components` property that could allow user to customize any component such as toolbar, search input, footer etc. 

Please star repo and wait for next releases. I will add this property.

tadeuqc:
Great! Thats awesome! If you need any help reach me out! I'm quite new to react but i've been doing a lot of stuff by myself lately!

mbrn:
@tadeuqc 

if you want to contribute you can look at the issues which i labeled as good first issue. You can contribute to resolving them. 

If you have any trouble i can help you.

mbrn:
Hi @tadeuqc ,

Sorry for late job:(. But this is done. You can change any component of material-table in next release. 

Next release will be published in a few days.

PR #109, opened 2018-12-09
Adds more options for rendering actions

chris-stormideas:
## Related Issue
None currently, can raise one if required

## Description
Adds more options to rendering row and selected actions.  Have added examples documentation.  Does not break/change existing behaviour.

There is a touch of duplicated code, can refactor to use a helper if preferred


mbrn:
Hi @chris-stormideas,

I have to remove these additionalt action properties. They are confusing. If developer has different cases he/she can use custom render feature and draw own buttons.

chris-stormideas:
Hi @mbrn 

No problem, we should be able to do what we need with #70 being added :+1:

mbrn:
Ok. I am working on it and i will publish it today as a new release.

Issue #64, opened 2018-10-25
Add Server side paging, filtering, sorting feature


mbrn:
Add Server side paging, filtering, sorting feature

jakeleventhal:
+1 any idea on when this will be completed?

mbrn:
Hi @jakeleventhal,

Nowadays i am working very hard for another project. But i will do this as soon as possible. Please star project and wait for next releases. 

Also why don't you fork this project and contribute it.

jakeleventhal:
I'm a very busy guy haha I can wait. Thanks for all the effort though.

plummer-flex:
@mbrn I see what you mean in regard to #174 to merge these issues?

IllusiveBagel:
This feature would be super useful for a project im doing. Any heads up on when this will be implemented?

IllusiveBagel:
ahh ok cool thanks this will be a good enough solution for now

vijthang61:
I have implemented server side implementation by adding custom handlers for pagesize, page change, searching, sorting, filtering. I can use that until the server side events are implemented as part of material table. Once available I can remove my custom code. For now I am able to manage without any issues

roseak:
Is there a way to use `customSort` to trigger a function that would reload the data so that sorting can be done server side? I can't seem to get it to do anything.  I'd like to click on one of the table headers and then have that call a function that makes a call with the table header as a param to sort server side.

mbrn:
@roseak , unfortunately for now.

vijthang61:
I have implemented server side sort by overriding the onOrderChange prop, I had written by custom implementation onOrderChange={this.handleOrderChange},
here I make a server side in the function handleOrderChange, which will update the table row state variable which will automatically render the UI.
In handleOrderChange function I update the a state variable for sort fields, once I set it it will immediately call ComponentDidUpdate method from where I make the ajax call and update the state variable for table row data. Once state variable is again updated it calls render again with new data. You have to check the old query and new query so that render, state variable update doesn't get called in infinite loop.

Sample code:
 /*
      Method which is invoked when the column sort order is changed
    */
    handleOrderChange = (orderedColumnId, orderDirection) => {        
      var columnField = this.props.columns[orderedColumnId].field;
      this.setState({sorting:{columnName: columnField, direction: orderDirection },loading: true});          
    };

queryString() {
      let queryUrl = `${URL}`;

      const { pageSize, currentPage } = this.state;
      const { filters } = this.state;
[...]

PR #137, opened 2018-12-31
add before & after event hooks which fixes material-table/issues/64

apemberton:
## Related Issue
#64 

## Description
As it stands, the onXyz events are triggered after data has already been loaded. By adding "before" events (and retaining the "after" events), we can easily enable remotely loading data before subsequently acting on it. I've bumped the version number to 1.12 and retained the original event handler names (eg: onChangePage) for backward compatibility (can just easily remove them). I can also add docs. I have a working example with react-redux and saga that remotely loads data from a REST API. Submitting this PR now a bit more bare bones to give you time to take a look and LMK what you think.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,6 +20,14 @@ export interface MaterialTableProps {
   onChangeRowsPerPage?: (pageSize: number) => void;
   onChangePage?: (page: number) => void;
   onOrderChange?: (orderBy: number, orderDirection: "asc" | "desc") => void;
+  onBeforeSelectionChange?: (data: any[]) => void;
+  onBeforeChangeRowsPerPage?: (pageSize: number) => void;
+  onBeforeChangePage?: (page: number) => void;
+  onBeforeOrderChange?: (orderBy: number, orderDirection: "asc" | "desc") => void;
+  onAfterSelectionChange?: (data: any[]) => void;
+  onAfterChangeRowsPerPage?: (pageSize: number) => void;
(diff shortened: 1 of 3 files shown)


mbrn:
Thanks for contribution @apemberton . I will review and approve it asap

mbrn:
I think resolve #64 with adding a HOC to material-table which override paging component. In overrided paging component we can handle all paging actions and manage data. 

Example:

```javascript
<MaterialTable
          columns={[...columns]}
          data={...data}
          components={{
            Pagination: props => <ServerSidePagination/>
          }}
        />
```
So we dont need before&after hooks of pagination. What do you think about this solution?

apemberton:
Providing custom pagination could be the right path, though the pagination and data fetching logic is in the MaterialTable and MTableBody (lines 35-40) components as well.

The existing pagination works fine though we do need to update the total count to the full total value from the server side and reset the current page always to 0 for the MTableBody component (since the pages are grabbed from the server side and we're always on page 0 so to speak) as follows:

```
<MaterialTable
  components={{
    Pagination: props => (
      <TablePagination {...props} count={results.total} />
    ),
    Body: props => (
      <MTableBody {...props} currentPage={0} />
    ),
  }}
  columns={columns}
  data={results.items}
/>
```

We actually don't need the before and after events - we just need the custom event handlers to fire before the setData methods are called. Given they fire after setData is called now, I thought it made sense to add before/after style handlers to preserve backward compatibility.

PR #172, opened 2019-01-21
Display a Loading Screen

Domino987:
## Description
This PR adds a circular progress bar to the table body to be displayed during loading of data.

- material-table
- m-table-body


mbrn:
I will apply a loading animation differently. So i won't accept this. But don't worry i will implement a new one at 1.14.0 release.

ArunSachdev:
> I will apply a loading animation differently. So i won't accept this. But don't worry i will implement a new one at 1.14.0 release.

Hey is there a way to push that loading animation upwards? when the table is too long the loading animation is not visible on screen and we have to scroll down for it.

Issue #174, opened 2019-01-22
Allow for external `count` prop


plummer-flex:
**Is your feature request related to a problem? Please describe.**
When handling filter/sort/pagination in external state, a user needs to be able to specify the total count of records in a set.

**Describe the solution you'd like**
A root level prop for `count` that, if set, will override the value passed to the `Pagination` component.

**Describe alternatives you've considered**
Overriding the `Pagination` component and injecting the `count` prop manually.

mbrn:
Duplicate with #64

PR #175, opened 2019-01-22
Add count as root prop

plummer-flex:
## Related Issue
#174

## Description
Add count global prop; if present override to Pagination

* src/material-table.js

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -26,6 +26,7 @@ class MaterialTable extends React.Component {
     }
     this.state = {
       columns: [],
+      count: calculatedProps.count,
       currentPage: 0,
       data: [],
       pageSize: calculatedProps.options.pageSize,
@@ -267,7 +268,7 @@ class MaterialTable extends React.Component {
               <props.components.Pagination
                 style={{ float: 'right' }}
                 colSpan={3}
-                count={this.state.renderData.length}
+                count={this.state.count || this.state.renderData.length}
                 rowsPerPage={this.state.pageSize}
                 rowsPerPageOptions={props.options.pageSizeOptions}
                 page={this.state.currentPage}
@@ -419,6 +420,7 @@ MaterialTable.defaultProps = {
   actions: [],
   classes: {},
   columns: [],
+  count: 0,
   components: {
     Actions: MTableActions,
     Body: MTableBody,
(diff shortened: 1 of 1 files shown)


mbrn:
Hi @plummer-flex ,

Why do you need this? If you add this feature to server side pagination, i will add complete solution for server side pagination, filtering etc.

plummer-flex:
Correct, this feature is needed for server-side pagination.

Currently, this is the only prop that can not be managed externally without overriding the `Pagination` component.  `onChangePage ` and `onChangeRowsPerPage` gets us most of the way there, but the pagination nav is still based on `renderData.length`.   Overriding that with an root prop is the simplest solution.

plummer-flex:
@mbrn I'm actually not following you.  Is there another implementation, or prop, for server-side pagination, filtering, or sorting?

mbrn:
I think you for other datatable implementations. I don't know other datatables' features.

vijthang61:
As mentioned by @plummer-flex , I would need this as well, the moment I override to make the count dynamic, the other props set like pagesize, onChangePage everything has to be provided as input to the custom pagination component again.

vijthang61:
components={{            
            Pagination: props => (             
              <TablePagination
                  rowsPerPageOptions={[5, 10, 25]}
                  colSpan={3}
                  count={rows.length}
                  rowsPerPage={rowsPerPage}
                  page={page}
                  SelectProps={{
                    native: true,
                  }}
                  onChangePage={this.handleChangePage}
                  onChangeRowsPerPage={this.handleChangeRowsPerPage}
                  ActionsComponent={TablePaginationActionsWrapped}
                />              
            ),
          }}
          onChangeRowsPerPage = {this.handleChangeRowsPerPage}
          onChangePage = {this.handleChangePage}

The last two lines become waste as I have to provide it as input to the custom pagination component again

plummer-flex:
@vijthang61 here's how I worked around it.
  - track local state for page, pageSize, column, direction
  - add handler functions for onChangePage, onChangeRowsPerPage, onOrderChange
  - override pagination and table body components to inject local state and settings
```
render() {
    const {
      columns,
      handlePage,
      handleRowsPerPage,
      handleOrder,
      handleSelect,
      icons
    } = this;

    const { page, pageSize } = this.state;

    const { count, data } = this.props;

    return (
      <MUITable
        title=""
        columns={columns}
        components={{
          Pagination: ({ ...otherProps }) => (
            <Pagination {...otherProps} icons={icons} count={count || 0} />
          ),
          Body: ({ renderData, ...subProps }) => (
            <Body
              {...subProps}
              currentPage={page}
              renderData={data}
              options={{ ...subProps.options, paging: false }} // set paging:false to avoid array slice
            />
          )
        }}
        data={data}
        onChangePage={handlePage}
        onChangeRowsPerPage={handleRowsPerPage}
        options={{ page, pageSize }}
        onOrderChange={handleOrder}
        onSelectionChange={handleSelect}
      />
    );
  }
```

balanka:
I am stile getting this error 
I am using material-table core Version  6.3.2

PR #177, opened 2019-01-23
Pagesize fix

Domino987:
## Description
If the page size is controlled, it does not get updated, if the prop changes after the initial mounting.
I added a catch for that in componentWillReceiveProps, so that if the prop changes, the state gets updated as well.

- material-table


mbrn:
How do you control page size? I didn't implemented controlled page size property.

Domino987:
Simply with options -> pagesize

mbrn:
Sorry i thougt about current page number:). I will apply this line to code.

Issue #181, opened 2019-01-25
Broken search functionality


netbrain:
```
    // Apply Search
    if (this.state && this.state.searchText) {
      renderData = renderData.filter(row => {
        let result = false;
        this.state.columns
          .filter(columnDef => { return columnDef.searchable === undefined ? !columnDef.hidden : columnDef.searchable })
          .forEach(columnDef => {
            if (columnDef.customFilterAndSearch) {
              result = !!columnDef.customFilterAndSearch(this.state.searchText, row, columnDef);
            }
            else if (columnDef.field) {
              const value = this.getFieldValue(row, columnDef);
              if (value && value.toString().toUpperCase().includes(this.state.searchText.toUpperCase())) {
                result = true;
              }
            }
          });

        return result;
      });
    }
```

The addition of the customFilterAndSearch functionality has partially broken the search functionality depending on where customFilterAndSearch is used. 

If for instance customFilterAndSearch was the last column in the column definition list (worst case) then only this field would be included in the search logic as it would overwrite any results from previous findings.

I'll be submitting a patch shortly.

PR #182, opened 2019-01-25
fixed issue 181

netbrain:
This PR fixes the issue #181 

and optimizes the search functionality slightly by using Array.some instead of Array.forEach. It's still O(n) though. But performance would still benefit greatly when iterating many rows and columns.


mbrn:
Code was changed after you opened this pr. But i took your code and commit it. Thanks @netbrain.

Issue #201, opened 2019-01-31
Error pagination and grouping if value is undefined


Domino987:
**Describe the bug**
If a column field is undefined of one object, and it is grouped by that field, it crashes.

**To Reproduce**
1. Go to: https://mbrn.github.io/material-table/#/docz-examples-12-example-grouping
2. Add items with one field undefined
3. Sort for that field

**Expected behavior**
It should not crash, if one field is undefined, but should use the empty label provided in the columns object

mbrn:
v1.17.0 published

PR #202, opened 2019-01-31
Fix this undefined for groupBy

Domino987:
Fix this undefined for groupBy

## Related Issue
#201 

## Description
Fixes the groupBy function to use arrow function, since this will be undefined in the reduce function. Arrow functions do not change this, so now it will not crash, if a value is undefined.

* material-table.js


mbrn:
I changed code as your code. Thanks @Domino987

PR #203, opened 2019-01-31
Added groupedBy to localization

Domino987:
## Description
Added grouped by localization to display different languages

* material-table
* group-bar
* index.d.ts


mbrn:
I applied all changed in this pr. Thanks @Domino987

PR #215, opened 2019-02-04
Prevent too much white space for empty rows

Domino987:
## Description
If the pagesize is e.g. 50, but only 10 items are added to the table, it creates ugly white space:
image
This PR prevents the empty rows to become to big (limits to 5) and looks now like this:
image

- m-table-body

### DIFF
--- a/src/m-table-body.js
+++ b/src/m-table-body.js
@@ -86,7 +86,7 @@ class MTableBody extends React.Component {
       const startIndex = this.props.currentPage * this.props.pageSize;
       const endIndex = startIndex + this.props.pageSize;
       renderData = renderData.slice(startIndex, endIndex);
-      emptyRowCount = this.props.pageSize - renderData.length;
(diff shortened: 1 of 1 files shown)


mbrn:
No need this @Domino987 

You can use emptyRowsWhenPaging: false in options

hsluoyz:
Hi @mbrn , I just found my table has too many empty rows and It takes me some time to find this issue. I think we should set ``emptyRowsWhenPaging`` to ``false`` by default, or at least mention it in the docs. Currently, the docs said nothing about this option.

Domino987:
I am sure @mbrn would love to see a PR for that 😉

mbrn:
Hi @hsluoyz 

I am working on a documentation boilerplate nowadays. (https://github.com/mbrn/libzy). Then i will rewrite a new documentation for material-table. It will be better, don't worry :)

PR #227, opened 2019-02-12
Toolbar title and button alignment

bobo-le:
## Description
It would be nice to disable the title in the toolbar and align the searchbar and buttons on the left.

* material-table
* m-table-toolbar


mbrn:
This is nice feature @bobo-le ,

But can you change freeButtonAlignment name as toolbarButtonAlignment. It could me more self-explanatory

bobo-le:
@mbrn yes sure, is done

PR #251, opened 2019-02-21
Disabled grouping for columns

### LINKED ISSUE #210: Grouping ignored render prop
**Describe the bug**
If a column is grouped, it ignores the render prop and if cannot render that value, it crashes.

**To Reproduce**
1. Go to https://mbrn.github.io/material-table/#/docz-examples-12-example-grouping
2. Change one field in data to be an object
3. Provide a render prop
4. Group that column

**Expected behavior**
The group row should render the render prop if provided

Domino987:
## Related Issue
#210 

## Description
This allows to disable the grouping of specific columns. This can be also used to fix #210. 

* MaterialTable
* TabeHeader


mbrn:
@Domino987 

We may need to change disableGrouping as grouping. Because we use same field name both for options an columns such as header style.

Domino987:
I changed the name to grouping as requested

PR #271, opened 2019-03-03
bug fix TypeError: Cannot read property isExpanded of undefined

heylenz:
bug fix TypeError: Cannot read property isExpanded of undefined


mbrn:
Sorry @heylenz. This is not appropriate solution. We should find why rowData is null.

mbrn:
I fixed this with 1.19.1

Issue #305, opened 2019-03-11
Stepped Pagination


Domino987:
Hi
I saw a stepped pagination feature in another table and thought that it might look good for this table as well.
I implemented a proof of concept and would create a PR after polishing, if you like it @mbrn ?
Here is a link to my implementation : [Stepped Pagination](https://github.com/Domino987/material-table/tree/SteppedPagination)
It would look like this:
![image](https://user-images.githubusercontent.com/17567991/54136217-a5eba380-441b-11e9-82bf-d41ad3e606c3.png)
![image](https://user-images.githubusercontent.com/17567991/54136264-b8fe7380-441b-11e9-9b56-0c81f5beb07b.png)
![image](https://user-images.githubusercontent.com/17567991/54136182-95d3c400-441b-11e9-85c2-f0b80d2b6fcb.png)
![image](https://user-images.githubusercontent.com/17567991/54136337-dfbcaa00-441b-11e9-8a9a-d801c79b6c58.png)

mbrn:
Easy and useful feature :) I will do

PR #330, opened 2019-03-15
Stepped pagination

Domino987:
## Related Issue
#305 

## Description
This shows the pagination with numbers and makes navigation overall easier.

- Material-Pagiantion
- index.d.ts
- docs


mbrn:
Hi @Domino987 

This is very nice contribution. But this should be an option. Can do these steps before merging:

1. Add another file as m-table-stepped-pagination.js. Copy your code in this file. 
2. Add paginationType field to options with 'normal' and 'stepped' options. 'normal' should be default value. 
3. Use appropriate pagination component in material-table.js 

So users could be select a pagination type.

Domino987:
Done :)

mbrn:
You are very fast :)

Issue #347, opened 2019-03-21
Removing select all button from table


malix1:
Hello.
How can I remove or disable select all button from table ?

mbrn:
Hi @malix1 ,

There is no any option to hide selectAll checkbox from header. If you want, i will be grateful with you PR. 

As a workaround you can override header component of table. Look at examples: 

https://mbrn.github.io/material-table/#/docz-examples-10-example-component-overriding

mbrn:
This is done with 1.26.1. 

Usage: 

```javascript
<MaterialTable
                options={{
                  selection: true,                  
showSelectAllCheckbox: false
                }}
............
              />
```

malix1:
Thank you
And i'm sorry. I could not help :(

Have a nice day

Mehmet Baran <notifications@github.com>, 21 Mar 2019 Per, 13:36 tarihinde
şunu yazdı:

mbrn:
Not important @malix1 , You give us a new simple feature idea :)

PR #349, opened 2019-03-21
Removing select all checkbox from table

malix1:
## Related Issue
#347 

## Description
Removing Select All checkbox from table

-----------------------------------------------
We can make a custom header component and send "displaySelectAll={false}" to it, it will not display the select all checkbox.


mbrn:
Thanks for contribution. But changes is not appropriate to project structure. I will do this feature asap. Thx again.

rchvalbo:
This was a super useful property. Why was it not merged in??

rchvalbo:
Nevermind, I see that it was implemented. Thank you!

PR #355, opened 2019-03-21
Add support for default `exportCsv` function.

### LINKED ISSUE #288: CSV: Export all data instead of current page
hi @mbrn 👋 

I'm trying to figure out, as the title says heheh, export all data to CSV instead of the current page, any tip on this?

I think a possible solution can add a prop to the public API that received a function that process all the data before the export, wdyt?

All the best

fernandes:
## Related Issue
This PR Closes #288

## Description
This PR add the `exportCsv` prop to material table, so the user is able to build it's own csv file.

* `MaterialTable`
* `MTableToolbar `

## Additional Notes
The solution is pretty much the same as proposed in the issue, I just needed to pass the `m-table-toolbar` so can call the `this.setState` and close the tooltip.

My suggestion would be "hook" the `defaultCsv` and close the tooltip inside the toolbar, so least exposure would be needed, but I'm keeping this PR simple, focused on the solution so we can discuss it better.

Thank youuu!! :D

## Example of Usage

I tested in my application and worked with the following code:

```jsx
import { CsvBuilder } from 'filefy';

function Index({model}) {
  const handleExportCsv = (toolbar, columns, renderData) => {
    const csvColumns = columns
      .filter(columnDef => {
        return !columnDef.hidden && columnDef.field && columnDef.export !== false; 
      });

    const data = renderData.map(rowData =>
      csvColumns.map(columnDef => rowData[columnDef.field])
    );

    const builder = new CsvBuilder(('foo') + '.csv')
      .setDelimeter(',')
      .setColumns(csvColumns.map(columnDef => columnDef.title))
      .addRows(data)
      .exportFile();

    toolbar.setState({ exportButtonAnchorEl: null }); // this is the reason to pass the m-table-toolbar object
  }

  return (
    <MaterialTable
      columns={[
        { title: 'ID', field: 'id' },
      ]}
      data={model}
      title="Title"
      options={{
        columnsButton: true,
        exportButton: true,
        actionsColumnIndex: -1,
        exportCsv: handleExportCsv,
        exportFileName: 'untitled.csv' // using custom this is not used anymore
      }}
    />
  );
}
```

### DIFF
--- a/docz/props.mdx
+++ b/docz/props.mdx
@@ -188,6 +188,7 @@ Options property could be given to component as `options` property. You can chan
 | exportButton               | boolean                | false         | Flag for export button that render export buttons                                 |
 | exportDelimiter            | string                 | ,             | Delimiter to use in exported CSV file                                             |
 | exportFileName             | string                 | Title         | Exported file name                                                                |
+| exportCsv                  | func                   |               | Function to create a custom CSV file                                              |
 | filtering                  | boolean                | false         | Flag for filtering row                                                            |
 | header                     | boolean                | true          | Flag for header visibility                                                        |
 | headerStyle                | object                 |               | Header cell style for all headers                                                 |
--- a/index.d.ts
+++ b/index.d.ts
@@ -145,6 +145,7 @@ export interface Options {
   exportButton?: boolean;
   exportDelimiter?: string;
   exportFileName?: string;
+  exportCsv?: (toolbar: MTableToolbar, columns: any[], renderData: any[]) => void;
   filtering?: boolean;
   header?: boolean;
   headerStyle?: React.CSSProperties;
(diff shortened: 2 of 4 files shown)


fernandes:
hi @mbrn , thanks for reply

Just fixed the 3 points you told, this is the first time I'm writing this typescript signature, not sure if I did the best choose, anything, please just tell me

thanks

revskill10:
Is there any way to make this more generic ? As the original library `filefy` allows more file types (CSV, PDF, Excel,...)

fernandes:
hi @revskill10 

actually filefy only supports CSV as stated on README

> Only CSV export is available for now!

but yeah, you can pass a custom function (that's the idea) that generates any kind of file you'd like

fernandes:
awesome, thank you @mbrn ! :)

waiting next release to use it! 😄

Issue #371, opened 2019-03-25
Select is missing localization?


TatiTheFreaK:
**Describe the bug**
In https://github.com/mbrn/material-table/blob/master/src/material-table.js , line 252 it looks like "rows" is hardcoded, so we can't localize it. Is this the case?

**Expected behavior**
Have a prop or something to allow us to localize the string

**Screenshots**
![image](https://user-images.githubusercontent.com/11199668/54946854-c6395900-4f17-11e9-932e-b1f900c516a2.png)

mbrn:
I developed it a few days ago. I will add localization for it asap. Or why you dont contribute: )

TatiTheFreaK:
It works now, thank you for the quick merge :D

PR #376, opened 2019-03-26
Add localization for label on select

TatiTheFreaK:
## Related Issue
#371 

## Description
Add localization for label on select

pagination


TatiTheFreaK:
Where should I put the proptypes, since I included the string inside the pagination object?

mbrn:
search for "MaterialTable.propTypes = {" in material-table.js

TatiTheFreaK:
image
I've put the prop inside the pagination object, so I wouldn't be able to insert the proptype here... Unless you want me to put it out of the object

mbrn:
Thanks for contribution @TatiTheFreaK . I will publish it asap.

Issue #398, opened 2019-04-01
exportCsv column with lookup data


ivakarlec:
**Describe the bug**
Data in column with lookup are exported with the lookup-keys, not the lookup-values.

**To Reproduce**
1. Define a column with the lookup feature
2. Export the table
3. In the exported table are in the column with lookup the lookup keys

**Expected behavior**
In the exported table should be the lookup values.

**Screenshots**
![lookup_export_csv](https://user-images.githubusercontent.com/4302545/55320361-2af43680-5477-11e9-9b1f-05635c42bf99.png)

**Desktop (please complete the following information):**
 - all OS, Browsers, Versions

**Smartphone (please complete the following information):**
 - all Devices, OS, Browsers, Versions

mbrn:
This is published with 1.28.3

PR #399, opened 2019-04-01
fix: exportCsv menu closing and lookup columns

ivakarlec:
## Related Issue
#398 

## Description
When a column has a lookup feature, than export the values of the lookup, not the keys.

## Related PRs
no related PRs

* m-table-toolbar


mbrn:
This is published with 1.28.3

Issue #417, opened 2019-04-04
MTableEditRow to use DataManager getFieldValue to support nested data fields


spazzarama:
**Is your feature request related to a problem? Please describe.**
Existing values for nested fields display as blank when editing a row.

**Describe the solution you'd like**
MTableEditRow to retrieve the current value of a cell using the data manager getFieldValue function as per MTableBodyRow

**Describe alternatives you've considered**
Copying data from sub-objects onto properties of the root record, however this gets cumbersome when interacting with a broader system.

**Additional context**
I have changes for a pull request to add support for this.

Related to #162

mbrn:
@spazzarama 

Please send your PR. I will merge it if it has not a side effect. I will examine.

PR #427, opened 2019-04-04
Use data manager getFieldValue for EditRow

spazzarama:
## Related Issue
#417

## Description
Use data manager getFieldValue to retrieve current value in EditRow

MTableBody and MTableEditRow


mbrn:
Hi @spazzarama,

Thanks for contribution but this may not work with columns that has lookup. Can you try it, please?

spazzarama:
@mbrn it works fine for lookups whether in a nested property or not. The data manager .getFieldValue already handles this.

mbrn:
@spazzarama ,

It didn't worked with columns that has lookup. But i resolved this problem with a solution. Asssume this pr merged :)

PR #433, opened 2019-04-07
Add support for Formik

truongtv22:
## Related Issue
Relate the Github issue with this PR using `#`

## Description
Simple words to describe the overall goals of the pull request's commits.

## Related PRs
List related PRs against other branches:

branch | PR
------ | ------
other_pr_production | [link]()
other_pr_master | [link]()

## Impacted Areas in Application
List general components of the application that this PR will affect:

*

## Additional Notes
This is optional, feel free to follow your hearth and write here :)

PR #435, opened 2019-04-07
Add custom inline edit components

### LINKED ISSUE #316: Add custom field component for inline edit.

ColinKrist:
## Related Issue
closes #316 

## Description
Expose edit cell components in the same manner that custom display cell rendering is done today via column definitions. 

* form-field.js - here we check for `renderEditCell` from the columnDef props and attempt to render it
* index.d.ts added `renderEditCell` types **NOTE:** I don't know TS very well. This was my best attempt at adding all the props send from `renderEditCell` to the user's custom component.
* m-table-edit-row.js edit to remove unused import left over from someone else's PR.
* props.mdx - updated to include the new definition. ripped from #438


gmichalec-pandora:
I've done pretty much the exact same thing in #438 , but I think yours is more complete. The only thing mine does that yours does not is update the docs to document the new column property. Do you want to do that? If so, then I will close my PR

ColinKrist:
updated PR, I prefer your naming convention so I switched it up. TypeScript definitions updated accordingly too.

mbrn:
Hi @ColinKrist 

I think that this is same with #433. I added feature that lets developer to override edit component for all columns or just one column.

ColinKrist:
@mbrn look forward to your contribution you mention in #433, I made this PR because #433 didn't have support for singling out certain cells and only could override the generic cell type, but it seems you plan to make another commit with a similar approach to what I had.

mbrn:
@ColinKrist 

If you mean overriding edit component just for one column i did it. Please look at 3r example at https://mbrn.github.io/material-table/#/docz-examples-15-example-editable

Issue #437, opened 2019-04-08
Allow customization of edit cell


gmichalec-pandora:
**Is your feature request related to a problem? Please describe.**
There are cases where the provided field types aren't sufficient. It would be nice to be able to customize the 'edit' mode of a particular field.

**Describe the solution you'd like**
Add a new column property to provide a function for a custom edit cell renderer.

**Describe alternatives you've considered**
Couldn't find any way around it.

**Additional context**
PR incoming

mbrn:
Hi @gmichalec-pandora 

Thanks for contribution. I will check it asap and merge it.

mbrn:
editComponent added.

PR #438, opened 2019-04-08
New column property for renderEditCell to allow customization of edit view

gmichalec-pandora:
## Related Issue
#437

## Description
This adds a new column property so that we can provide a custom render function for the edit mode view of a field

* FormField


gmichalec-pandora:
Made redundant by #435

Issue #439, opened 2019-04-08
Move editRow to be the first row in table


gmichalec-pandora:
**Is your feature request related to a problem? Please describe.**
The current 'Add row' functionality renders the new row interface as the last row in the table. If you have a large table, this may be off the screen, leaving the user confused/unable to find the row edit interface.

**Describe the solution you'd like**
By moving the new row interface to be the first row in the table, it is much more discoverable - immediately below the add row button the was clicked.

**Describe alternatives you've considered**
You could do something like auto-scrolling the table perhaps, but this seems simpler and more intuitive. I also experimented with rendering the add row interface outside the table (below the header), but this solution works better.

**Additional context**
![material-table](https://user-images.githubusercontent.com/39168929/55750136-85e9d880-59f7-11e9-999f-32bcadc74138.png)

ColinKrist:
omg yes

mbrn:
editRowPosition feature added with version 1.31.0

SarthakU:
This is addRowPosition in options now, if anyone arrives here from google.

SarthakU:
This is addRowPosition in options now, if anyone arrives here from google.

awarmfastbear:
code example:


```
<MaterialTable
....
options={{
addRowPosition: "first"
}}
...
/>

PR #440, opened 2019-04-08
Edit row at top

gmichalec-pandora:
## Related Issue
 #439

## Description
Just moves the render of the edit row to the top of the table, below the filters.

* TableBody


ColinKrist:
I also just wrote a PR for this yesterday #435 , can you update the TypeScript definitions for your function?
Something like this is what is needed: https://github.com/mbrn/material-table/pull/435/commits/1c7a38f9b5450912cd25a42febb80ef40e6768a6#diff-b52768974e6bc0faccb7d4b75b162c99R61

ColinKrist:
Title of this PR should probably change to reflect the two things it is trying to solve:
1. top row edit
2. custom inline cells

gmichalec-pandora:
oops - i forgot made this branch off the wrong trunk - this was supposed to be standalone. I'll kill this one and re-open a new one. i have another PR for the edit cells feature - I'll comment about that one there.

PR #441, opened 2019-04-08
move addrow to be the first table row

gmichalec-pandora:
## Related Issue
 #439

## Description
Just moves the render of the edit row to the top of the table, below the filters.

* TableBody


mbrn:
Hi @gmichalec-pandora 

I added a addRowPosition feature to options. So it could be changed as `first` or `last`

gmichalec-pandora:
This is great - thanks! Amazing work!

sabahath786:
@mbrn I added a addRowPosition feature to options. So it could be changed as first or last, which is in current implmentation adn it works really great !!
What if i want to add row in between the rows with additional action icon on each row? 

I appreciate you help as soon as possible.
Thanks in advance.

attachedGithub

PR #481, opened 2019-04-18
Add option for exporting all data instead of just the visible rows

laitalaj:
## Related Issue
None

## Description
At the moment, material-table's CSV export only exports the data that is currently visible. This change makes it possible to instead export all data by passing `exportAllData: true` in the options.

## Related PRs
None

* CSV exporting

## Additional Notes
material-table is the best batteries-included table package for material UI that we've encountered so far. Thanks for all your hard work!


czsyjss:
@laitalaj The developer has added exportAllData Field. Plz check https://material-table.com/#/docs/all-props

PR #488, opened 2019-04-20
Implemented sticky-columns

hiroshifariassaito:
## Description
Enable left or right sticky column through a property

* Table cell
* Table filter row
* Table header

## Additional Notes

Peek 2019-04-20 01-10

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -95,10 +95,12 @@ export interface Column {
   grouping?: boolean;
   headerStyle?: React.CSSProperties;
   hidden?: boolean;
+  leftSticky?: boolean;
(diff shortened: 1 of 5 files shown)


mbrn:
Hi @hiroshifariassaito 

Thanks for very nice feature. But what happens when i set first two columns as left sticky?

hiroshifariassaito:
Well thought, we would have to set a width, and then set the `left`property. I don't see a viable way to do this right now.

mbrn:
Yes i want to add this feature for multiple columns. If i accept this pr now we can not change it never. But thanks for contribution again.

Issue #500, opened 2019-04-22
render fires even no changes on data source


devel0:
**Describe the bug**
Changing a textfield outside the material table cause column render fire.

**To Reproduce**
https://codesandbox.io/s/github/devel0/repros/tree/2f37d7e3d240d36791ae24572e441c08f1ac35c7/material-table/test01

**Expected behavior**
rerender should limited to data source

**Screenshots**
![Peek 2019-04-22 18-08](https://user-images.githubusercontent.com/13405008/56510284-9d61af00-6529-11e9-9a66-cdf2d85c5a82.gif)

render count incremented by the render of the column not from the text input itself ( [code](https://github.com/devel0/repros/blob/2f37d7e3d240d36791ae24572e441c08f1ac35c7/material-table/test01/src/App.tsx#L64-L67) )

mbrn:
Hi @devel0 

Actually i didn't worked on performance of material-table. But i know it has some performance issues. I will examine them asap. 

Thanks

devel0:
Hi @mbrn,
I wondering if render shouldn't be controlled through a table prop ( a number ) that user can connect to a state variable, so that table render proceed only on internal events ( page changed, etc ) or when this prop "age" changes respect to previous received values.
For example I can increment that number in my app when I know datasource changed and this way material table know that data source changed and allow render or not; maybe the way is to use that in `ShouldComponentUpdate`.
I don't know if there are other ways which the table can observe a generic dataset in an efficient way.

mbrn:
Thanks for recommendation. I will consider this way when working on performance.

masbaehr:
The problem becomes very serious especially if you have a "render" method in your columns which is doing some longer-lasting work. Then it can happen that the UI blocks completely.

PR #507, opened 2019-04-24
Data age

devel0:
## Related Issue
Relate the Github issue with this PR using #500 

## Description
dataAge table prop if omitted works in a backward compatible way
if specified a value more than -1 it can be used to communicate the table that datasource changed causing render to be accepted

- render performance

## Additional Notes
follow [sandbox](https://codesandbox.io/s/github/devel0/repros/tree/b0668933257b011f03bfbe34115476b9e78a3ac5/material-table/test01) shows how this works in a backward mode and then using dataAge to fire render on datasource changes

Peek 2019-04-24 19-08x

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,6 +7,7 @@ export interface MaterialTableProps {
   columns: Column[];
   components?: Components;
   data: any[] | ((query: Query) => Promise<QueryResult>);
+  dataAge?: number;
   detailPanel?: ((rowData: any) => React.ReactNode) | (DetailPanel | ((rowData: any) => DetailPanel))[];
   editable?: {
     onRowAdd?: (newData: any) => Promise<void>;
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -51,7 +51,7 @@ class MaterialTable extends React.Component {
     };
   }
 
-  componentDidMount() {
+  componentDidMount() {    
     this.setState(this.dataManager.getRenderState(), () => {
       if (this.isRemoteData()) {
         this.onQueryChange(this.state.query);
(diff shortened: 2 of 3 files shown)


devel0:
there is a minor issue that cause table not refresh first time `dataAge` is set; to solve that just increase dataAge twice and table refresh instantly; it may related to the react lifecycle render, shouldcomponent update, etc.

At the moment I using this PR in a project where dataAge is attached to a redux prop; updating dataAge after item subsequently to any save or items loaded make it work as expected with no more refresh if changing state for other controls external to the table.

devel0:
I converted invalidated from boolean to a number in order to allow searchText render correctly; without it it lost digited chars

mbrn:
Hi @devel0 

Thanks for contribution. But I am not sure, this is the effective way to solve this problem. I will take a look at your solution again. I am merging prs by order.

devel0:
Hi @mbrn ,
Yes the solution provided with PR is more a workaround but I pulled just to let you know one of possible approach and to have an alternative when a better solution coming up to have a comparision.
Btw, I'm not very experienced in js so I could wrong, but I think that the use of a support variable to allow user communicate that datasource changed should a feasible way. Another way like in c# could be using an Observable Collection ( here a [package](https://www.npmjs.com/package/observable-collection) ) to provide the datasource so that material table if states ds instanceof that can listen to datasource changes to "invalidate" the view automatically without having the user to worry about, but I don't know if such solution could results in a poor performance due to the fact that the observable collection should track changes and more the user should have to enclose its datasource in that and this would be framework constraint addition.
So definitively a solution with something similar to dataAge prop but written more appropriately ( using dataManager ) could not be so a bad approach because at the end the developer knows points where its datasource changes ( on load, on add, delete, update rows activities ) and working on that property shouldn't difficult ( personally I use redux from a global store ).
Thanks you too and others that contributed to this nice component, it's helpful to me.

mbrn:
We can use shouldComponentUpdate instead of a increasing a age variable.

PR #520, opened 2019-04-27
This closes #519, cellStyle takes rowData

### LINKED ISSUE #519: cellStyle should receive the rowData as second parameter
Sometimes the cell style is not determined by the cell data and needs to be checked how it relates to sibling columns.

For instance if I have 2 columns: `type` and `value`, I may highlight the `value` according to thresholds defined for `type`.

`cellStyle` could pass the full `rowData` as second parameter and make it possible :-)

barbieri:
add `rowData` as second parameter to `cellStyle` if it's a function.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -78,7 +78,7 @@ export interface EditCellColumnDef {
 }
 
 export interface Column {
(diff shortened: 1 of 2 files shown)


mbrn:
Thanks for contribution @barbieri

PR #525, opened 2019-04-29
This closes #474, lookup filtering returns empty results

### LINKED ISSUE #474: When using lookup the filtering returns empty results
**Describe the bug**
When using the filter using lookup, it returns "No records to display"

**To Reproduce**
1. Go to https://mbrn.github.io/material-table/#/docz-examples-10-example-component-overriding
2. Select the filter with option 'Şanlıurfa' on the column 'Doğum Yeri'
4. See 'No records to display'

**Expected behavior**
Data records with value 'Şanlıurfa' in the  'Doğum Yeri' field should be seen in the table

**Screenshots**
Print 1
Print 2

 - OS: Windows 10
 - Browser chrome 73.0.3683.103 64 bits

**Additional context**
Looking at the code, I checked that it takes the text from the table, and matches the value, so it returns empty.
`material-table\dist\utils\data-manager.js`
material-table\dist\utils\data-manager.js
Print 4

john-pc:
## Related Issue
#474 

## Description
lookup filter didn't worked on values 
added filter value flag for getFieldValue function in data-manager

* utils/data-manager.js

## Additional Notes
❤️

### DIFF
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -332,9 +332,9 @@ export default class DataManager {
     return node;
   }
 
-  getFieldValue = (rowData, columnDef) => {
+  getFieldValue = (rowData, columnDef, inFilter) => {
     let value = (typeof rowData[columnDef.field] !== 'undefined' ? rowData[columnDef.field] : byString(rowData, columnDef.field));
-    if (columnDef.lookup) {
+    if (columnDef.lookup && !inFilter) {
       value = columnDef.lookup[value];
     }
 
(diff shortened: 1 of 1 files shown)


src/utils/data-manager.js
```diff
@@ -332,9 +332,9 @@ export default class DataManager {
     return node;
   }
 
-  getFieldValue = (rowData, columnDef) => {
+  getFieldValue = (rowData, columnDef, inFilter) => {
```
mbrn: can you change it as (rowData, columnDef, lookup = true) , please? Because it may be used for different cases.

mbrn:
I did it. It will be on next release.

PR #530, opened 2019-05-01
Allow grouping by fields with render functions

Foosballfan:
## Description
I have cells which contain objects, e.g. { latitude: 0.5, longitude: 0.5 }. 
Everything displays as expected when using a render function on the column e.g. {render:  (data) => data.coords.latitude }

However, when grouping the render function is ignored and the groupData() function tries to group the [Object Object].

The column render function is currently used further down the  render chain in the table-group-row.

In this PR, I propose moving the render() call higher up the chain to the groupData() function so that grouping works correctly while still displaying correctly in the component render.

* Grouping

### DIFF
--- a/src/m-table-group-row.js
+++ b/src/m-table-group-row.js
@@ -79,10 +79,7 @@ export default class MTableGroupRow extends React.Component {
     }
 
     let value = this.props.groupData.value;
-    if(column.render) {
-      value = column.render(value, 'group');
-    }
-    else if (column.lookup) {
+    if (column.lookup) {
       value = column.lookup[value];
     }
     if((value === undefined || value === null) && column.emptyValue) {
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -566,7 +566,7 @@ export default class DataManager {
 
       let object = result;
       object = groups.reduce((o, colDef) => {
-        const value = current[colDef.field] || byString(current, colDef.field);
+        const value = colDef.render ? colDef.render(current) : (current[colDef.field] || byString(current, colDef.field));
         let group = o.groups.find(g => g.value === value);
         if (!group) {
           const path = [...(o.path || []), value];
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -566,7 +566,10 @@ export default class DataManager {
 
       let object = result;
       object = groups.reduce((o, colDef) => {
-        const value = colDef.render ? colDef.render(current) : (current[colDef.field] || byString(current, colDef.field));
+        const value = 
+          current[colDef.field] 
+            ? ( colDef.render ? colDef.render(current[colDef.field], 'group') : current[colDef.field] )
+            : byString(current, colDef.field);
         let group = o.groups.find(g => g.value === value);
         if (!group) {
           const path = [...(o.path || []), value];


src/m-table-group-row.js
```diff
@@ -79,10 +79,7 @@ export default class MTableGroupRow extends React.Component {
     }
 
     let value = this.props.groupData.value;
-    if(column.render) {
-      value = column.render(value, 'group');
-    }
-    else if (column.lookup) {
+    if (column.lookup) {
```
Foosballfan: Here the render call is no longer required as groupData() has already run it.

src/utils/data-manager.js
```diff
@@ -566,7 +566,7 @@ export default class DataManager {
 
       let object = result;
       object = groups.reduce((o, colDef) => {
-        const value = current[colDef.field] || byString(current, colDef.field);
+        const value = colDef.render ? colDef.render(current) : (current[colDef.field] || byString(current, colDef.field));
```
Foosballfan: If there is a render function on the column then execute it to get the group-by value

mbrn: But what happens when render returns and react element?

Foosballfan: That's a good point, I haven't considered that here. 

At the risk of bloating the library, my first thought is to introduce a 'dataAccessor' on the column which could then be used to return primitive values.

What are your thoughts on the best way to handle this scenario?

mbrn: 1. field must be filled if we want to group data. 
2. Grouping will work on data[field]
3. render works properly when without grouping.
4. render won't work when column grouped. Because render requires all rowData to work. But we are on a group, not row. We may call render with data[field] instead of all row data. So user must handle this case.

Foosballfan: I understand each of those points but I'm not sure what you're trying to say there. Can you elaborate?

mbrn: I mean important point is calling render method for a grouped row. First of all, we should find a good solution for it.

Foosballfan: OK, so at the moment there is a 'type' of group or row which is passed in to the render which allows you to deal with the difference in data type.

Perhaps it would be a neater solution to split it into two props, renderRow and renderGroup. 
That way we can specify through docs and typescript that renderRow may return a component but renderGroup must return a primitive, in order for the groupData() function to work.

Does that sound good to you?

mbrn: No. field already did this. We can call render method with an additional parameter ('row' | 'group') to make a difference in render method according to group or row render.

Foosballfan: Yeah, absolutely. 

The only issue I see is type safety to ensure that people know they have to be returning a primitive when it is a group.

Is that something that you are happy to leave as just documentation?

mbrn: Yes you can add a blockquote to grouping page on documentation. 

Do you know new documentation is on another repo (github.com/mbrn/material-table.com)

Foosballfan: Ok, great. Let me make the changes here and if it looks good to you then I'll update the docs to reflect it.

src/utils/data-manager.js
```diff
@@ -566,7 +566,10 @@ export default class DataManager {
 
       let object = result;
       object = groups.reduce((o, colDef) => {
-        const value = current[colDef.field] || byString(current, colDef.field);
+        const value =
```
Foosballfan: @mbrn Is this what you were looking for?

Foosballfan:
@mbrn have you had a chance to look at this? I will update the docs once you are happy with the logic.

Foosballfan:
@mbrn Would really like to make progress with this

mbrn:
Hi @Foosballfan ,

Sorry for late response. I worked on this. Can you check latest version. It should work for columns that has a render.

Foosballfan:
No longer having issues with latest release

Foosballfan:
> Hi @Foosballfan ,
> 
> Sorry for late response. I worked on this. Can you check latest version. It should work for columns that has a render.

Hi @mbrn, I understand you must be very busy. Thanks for looking into this.

crewtaylor:
I am getting errors in 1.39 when I try to group a column that has a custom render function. It is stating that the rowData for the column has a null value.

Foosballfan:
The data passed in to the render function is different if it is for a group. There is a second argument passed in to render with the type. You need to handle it with a conditional.

PR #575, opened 2019-05-13
Group key null check

Foosballfan:
## Description
When a value equals 0, the existing logic for the key of a Group Row can cause key errors and thus repeated rows.
This change does not count 0 as a null value.

*Grouping

Screenshot 2019-05-13 at 12 29 18


mbrn:
No difference

Foosballfan:
@mbrn can you please elaborate? I went to a lot of effort describing the change and your 2 word response is very unclear.

mbrn:
Hi @Foosballfan,

Thanks for your effort. But i asked a question(What is the difference?).

Foosballfan:
The difference is that 0 was being treated as a null value, when it is in fact a valid value.

The change I made is to only treat null or undefined as null values.

In the screen shot I show how treating 0 as a null value causes the rows to be duplicated as a result of React duplicate key errors.

> When a value equals 0, the existing logic for the key of a Group Row can cause key errors and thus repeated rows.
>This change does not count 0 as a null value.

mbrn:
This is the answer. Thanks.

Issue #549, opened 2019-05-08
Is it able to move the detail panel `Icon` from left to right


YongZhangEmbed:
![image](https://user-images.githubusercontent.com/46992381/57343440-f5580280-7132-11e9-948e-90f3c2f7fea9.png)

mbrn:
Currently no. But you can use onRowClick to toggle detail panel. 

Look at 3rd example at:
https://material-table.com/#/docs/features/detail-panel

YongZhangEmbed:
Is it possible to remove the `Icon`?

mbrn:
You can not. But you i can add this feature. Because a lot of developer asked for it.

YongZhangEmbed:
thanks for your feedback, hope this feature coming soon

Leocete:
@mbrn, has the feature of hiding the "toggle" icon been implemented?
Also, is it possible to show a detail panel by default (without clicking anywhere)?

edwinckc:
> @mbrn, has the feature of hiding the "toggle" icon been implemented?
> Also, is it possible to show a detail panel by default (without clicking anywhere)?

Not related to the original issue, but I was also looking for this feature. Using this as a workaround:

```
detailPanel={[
    (rowData) => ({
        disabled: isEmpty(rowData.notes),  // hides hover state
        icon: isEmpty(rowData.notes) ? ' ' : ChatBubbleOutline,  // single space used here, empty string defaults to chevron
        openIcon: isEmpty(rowData.notes) ? ' ' : ChatBubble, // single space
        ...
    }),
]}
```

yzhaoa:
Try 

```
      sx={{
        '.MuiDataGrid-row > .MuiDataGrid-cell:first-child, * > .MuiDataGrid-columnHeader:first-child': {
          order: 1,
        }
      }}

```

PR #587, opened 2019-05-16
Feat: able to align the detail panel icon column to right

EbenZhangEmbed:
## Related Issue
#549

## Description
Able to align the detail panel columns to the right hand side of the table

* Table header
* Table body
* Detail panel columns


michaelgthomas:
Thank you for this, you saved me a lot of heartache tonight.

PR #595, opened 2019-05-17
Added label functionality to actions

phishfordead:
## Description
Allows the use of a label with actions. The label can either be a component in which the component will be loaded below the icon, or can be a string, in which case it will be loaded inside a div below the icon.

branch 
feature/labels-on-actions

*m-table-actions.js (added the ability to have a text label)
*prop-types.js (added label as a property of action)
*demo.js (added a test action button)

## Additional Notes
A few things, a npm install did not work when I forked the repo. I had to hand install all dependencies. Also, I added m-table-actions.css to change the class MuiIconButton-label-* to have flex-wrap: wrap to allow the label to be positioned beneath the Icon for the label.  In additon to adding the css file I had to install style-loader and css-loader from npm.  I updated the webpack.config.js file of the demo application to reflect these additions.


mbrn:
Hi @phishfordead ,

Thanks for your effort. But i want to add label functionality for normal buttons instead of div. I will implement it asap.

tschaka1904:
How would this work then? I mean, if you want a freeAction with a label?

Issue #192, opened 2019-01-30
Filter Usability


jlgreene2:
After using your control for a little while now, there are some aspects of how filters are handled that I'm wondering about whether they could be implemented in a different way.  

I believe that having the filter input fields show up persistently on a separate header line of the Grid takes up a lot of screen real estate.  I'm wondering if it might be more useful to render the Filter icon to the far right hand side of the Header column next to the header label, and always render a dropdown that includes a text entry field for manually entering filters, as well as any custom ones that have been defined as the result of including a lookup property for the column.  

You could change the color of the filter icon when filters are applied, rather than listing all of the selected filters as is done today.  The current implementation does some strange resizing of the column when multiple long filter lookup items are selected.  This could be avoided by simply always providing a dropdown component for entering/managing filter values.

Also, when you have a column that has a lot of lookup values, it is not visually apparent that you can scroll down to see additional filter values that do not currently fit within the view of the filter dropdown. I think it might be better to set a minimum size of the filter dropdown that is a little larger, or try to calculate the height based on the number of lookup values specified, with a break at some defined percentage less than the remaining height of the screen.

This is a great component.  I'm really enjoying using it in my application.

MaxGorshkov:
thank you for great component!

If you improve filter (like a google docs excel), it will be absolutely best)

![image](https://user-images.githubusercontent.com/14037804/57358495-fd459000-718e-11e9-80ab-2cdd540e0e10.png)

mbrn:
I have plans for a new filtering component but now i am focused on stability and performance. Then i can work for new features

PR #614, opened 2019-05-23
Adding filter to header row

PavelS111:
## Related Issue
 #192

## Description
New Feature: showing filter either in header or in separate row.
The color of the filter icon is changing to dark when filters are applied.

`filterType: PropTypes.oneOf(['row', 'header']),`

image


mbrn:
Hi @PavelS111 ,

Thanks for your effort. But this is not in my roadmap. We already have a filter row. I will add a filter bar(Chips) as a second filtering option.

scipe:
Hi dear @mbrn: just wanna ask about your last message here: Is chips already in roadmap?

PR #652, opened 2019-05-31
Add aggregation support for grouping by column

paillave:
## Description
Permits to add a new grouping header line that show the result of aggregations if they are setup for the matching column

Only the grouping feature is impacted. It should have a very low impact if no column has an aggregation that is setup.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .docz
-.vscode
 dist
 node_modules
 package-lock.json
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,7 @@
             "request": "launch",
             "name": "Launch Chrome against localhost",
             "url": "http://localhost:8080",
-            "webRoot": "${workspaceFolder}",
+            "webRoot": "${workspaceFolder}/demo",
             "sourceMaps": true,
             // "preLaunchTask": "npm: start"
         }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,7 +11,7 @@
             "url": "http://localhost:8080",
             "webRoot": "${workspaceFolder}",
             "sourceMaps": true,
-            "preLaunchTask": "npm: start"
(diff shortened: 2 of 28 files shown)


paillave:
@mbrn Any issue with this pull request?
Shall I make any corrections?

mbrn:
Actually i didn't checked it yet. It contains many changes on files.

paillave:
no pb, if any question or help needed, let me know.

paillave:
don't accept this pull request as is, githut automatically includes pushes I made on my personal fork after I made this pull request... I'll ensure it is ok tonight.

paillave:
I reverted a temporary change I made on my personal fork that github automatically included in this pull request afterward (didn't know it worked this way)

paillave:
I can remove the config file (it was to work on my branch as I work on several computers).
What dead code (obsolete code that is never called) are you talking about?

mbrn:
Hi @paillave ,

I have checked this pr and really like it. But i will try to another way to implement summary row and group aggregation together. So i won't merge this code but i will copy some piece of it. Thanks for your great effort.

paillave:
Hi, no worries. I tried to find a way so that, ergonomically speaking, we could have both summary and aggregation on the same line (would look nicer), but a separate line seemed to be the only solution not break the entire existing code regarding columns layout of the component. Any idea of when you could spend some time on this feature? :)

mbrn:
This feature has most top priority. But firstly i will focus to reduce number of bug issues.

paillave:
After reviewing devxtreme data grid, I see how to resolve the visual pb.
Actually, in tree mode, they assume that the first column will never move as it is some sort of business identifier (like the name of the business object that is listed). Therefore they indent the content of the column within the cell. Their grid doesn't offer any aggregation feature but this way of doing things would permit to solve two problems:
* rendering properly the aggregation line with the name
* having a good looking tree if even if there is no selection checkbox rendered

PR #722, opened 2019-06-14
Tables now take an option to enable dense padding, which increases information density

gianlucamateo:

* Tables and Tree views now have a dense padding mode that can be activated in the options props

## Additional Notes
I did not get to test these changes against all features, I tried actions and tree views with selectors, which both work. I'm unsure on the consequences on detail views or other features


gianlucamateo:
There seems to be an issue with the propTypes, of which I have no knowledge. I would appreciate a bit of help

masbaehr:
Isn't there a CSS-Only way to make the table look narrow? For example i tried to remove the padding from the cells and the icons.. without any code change.. 

But i'd like to know if i can achieve this without overwriting the actual CSS classes manually. (Only with options)

gianlucamateo:
Of course you could just override css styles for everything. But material-table is based on MUI, and components like the IconButton take properties that decide size. Passing a prop and propagating it to the MUI Components that accept it is the clean way to do it IMO, and less prone to breaking with updates

mbrn:
Hi @gianlucamateo,

Thanks for contribution. I could not find any problem with code.

PR #731, opened 2019-06-17
Add the possibility add placeholder on filtering

AugustoCalaca:
## Description
Now the user can pass a prop in element of the column named 'filterPlaceholder' and the value that he wants
Ex: 
_column: {[
  { title: 'Id', field: 'id', **filterPlaceholder: 'placeholder'** },
]}_

Afects the component **m-table-filter-row** only on function **renderDefaultFilter**

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -124,7 +124,7 @@ class App extends Component {
                     />
                   ),
                 },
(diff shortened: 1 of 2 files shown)


mbrn:
Thanks for contribution

Issue #601, opened 2019-05-20
Add custom styles


pwork2015:
Hi!
Please add option customStyles here:
![изображение](https://user-images.githubusercontent.com/50596432/57999805-5f7f8880-7adf-11e9-9820-731583df8923.png)
style: {
  maxHeight: props.options.maxBodyHeight,
  overflowY: 'auto',
  ...(props.options.customStyles || {}),
 }

zproxy00:
Cannot find changes, and "height" doesn't work.
<props.components.Container style={{ position: 'relative', ...props.style }}> - that?
It works not as I need. Need to add properties to style: { maxHeight, overflowY, ...HERE... } as in my screenshot.

pwork2015:
Cannot find changes, and "height" doesn't work.
<props.components.Container style={{ position: 'relative', ...props.style }}> - that?
It works not as I need. Need to add properties to style: { maxHeight, overflowY, ...HERE... } as in my screenshot.

mbrn:
This feautere did not published yet. It will be in next release(tomorrow)

mbrn:
This is done with version 1.38.0

wfaucet:
Cannot understand how to use it. Where to set style.height, for example? You did not add code where I ask.

pwork2015:
Cannot understand how to use it. Where to set style.height, for example? You did not add code where I ask.

mbrn:
<MaterialTable
      style={{padding: 20}}
...
/>

pwork2015:
It works not as we need. I still cannot set style of inner div.
This

style: {
     maxHeight: props.options.maxBodyHeight,
     overflowY: 'auto'
 }

is not editable, but it must be editable..

pwork2015:
where can I find this part in source code (/src )?

![изображение](https://user-images.githubusercontent.com/50596432/57999805-5f7f8880-7adf-11e9-9820-731583df8923.png)

PR #734, opened 2019-06-18
Add body style

pwork2015:
## Related Issue
https://github.com/mbrn/material-table/issues/601

## Description
Add custom body style


mbrn:
Hi @pwork2015 ,

Thanks for contribution but i will add jss classes support for MT asap. So this would be unuseful. Thanks again.

PR #738, opened 2019-06-19
Change page if it is out of range

### LINKED ISSUE #708: Removing only item from last page leaves user out of page range
**Describe the bug**
Deleting the only item in a 10 page table leaves the user on the 10th page rather than displaying the 9th page. The pagination also reads '46-45 of 45' and 'No records to display' is visible

**To Reproduce**
1.Create multiple paged table
2. Delete last and only element
3. Observe table says 'No records to display'

**Expected behavior**
The Table should load the previous page which contains data

**Screenshots**
image
image

 - OS: Ubuntu 16.04
 - Browser Chrome

**Additional context**
Happy to put a PR in, two options from quickly skimming the code seem to be;
- Call changePage from MaterialTable component in componentWillReceiveProps
- Call changePage in Pagination component

AndySeymour2904:
## Related Issue
Fixes #708 

## Description
Change the page if it is out of range from either deleting rows or filtering

## Related PRs
N/A

MaterialTable

## Additional Notes
The error still appears in the console as we don't catch it before the update.

The ideal solution would be to use getDerivedStateFromProps and also use it to replace `UNSAFE_componentWillReceiveProps`. This seemed like quite a major refactor though, so I decided against it

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -85,12 +85,12 @@ export default class MaterialTable extends React.Component {
   }
 
   componentDidUpdate() {
-    const count = this.isRemoteData() ? this.state.query.totalCount : this.state.data.length
-    const currentPage = this.isRemoteData() ? this.state.query.page : this.state.currentPage
-    const pageSize = this.isRemoteData() ? this.state.query.pageSize : this.state.pageSize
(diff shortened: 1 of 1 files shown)


mbrn:
Thanks for contribution

Issue #750, opened 2019-06-24
Search functionality broken by pagination


makr11:
If I search for an item on page X (e.g. 32) and number of items are below page 32, pagination is left on page 32 and warning is shown (Warning: Failed prop type: Material-UI: the page prop of a TablePagination is out of range (0 to 6, but page is 32)).

![paginationBug](https://user-images.githubusercontent.com/31960173/60014420-45f2d300-9681-11e9-8a9e-bf50bcdfe8fc.png)

gianlucamateo:
should be resolved by PR #754

vinterdo:
Issue is still present on version 1.45.0. Same steps to reproduce as described above.

Additionaly, it causes Invariant Violation: Maximum update depth exceeded later:
`
in the <span> component:
    in span (created by Tooltip)
    in Tooltip (created by WithStyles(Tooltip))
    in WithStyles(Tooltip) (created by MTablePaginationInner)
    in div (created by MTablePaginationInner)
    in MTablePaginationInner (created by WithStyles(MTablePaginationInner))
    in WithStyles(MTablePaginationInner) (created by ActionsComponent)
    in ActionsComponent (created by ForwardRef(TablePagination))
    in div (created by ForwardRef(Toolbar))
    in ForwardRef(Toolbar) (created by WithStyles(ForwardRef(Toolbar)))
    in WithStyles(ForwardRef(Toolbar)) (created by ForwardRef(TablePagination))
    in td (created by ForwardRef(TableCell))
    in ForwardRef(TableCell) (created by WithStyles(ForwardRef(TableCell)))
    in WithStyles(ForwardRef(TableCell)) (created by ForwardRef(TablePagination))
    in ForwardRef(TablePagination) (created by WithStyles(ForwardRef(TablePagination)))
    in WithStyles(ForwardRef(TablePagination)) (created by MaterialTable)
    in tr (created by ForwardRef(TableRow))
    in ForwardRef(TableRow) (created by WithStyles(ForwardRef(TableRow)))
    in WithStyles(ForwardRef(TableRow)) (created by MaterialTable)
    in tfoot (created by ForwardRef(TableFooter))
    in ForwardRef(TableFooter) (created by WithStyles(ForwardRef(TableFooter)))
    in WithStyles(ForwardRef(TableFooter)) (created by MaterialTable)
    in table (created by ForwardRef(Table))
    in ForwardRef(Table) (created by WithStyles(ForwardRef(Table)))
[...]

PR #754, opened 2019-06-25
reset page to 0 when searching

gianlucamateo:
## Related Issue
#750 

## Description
resets page to 0 when changing search text


mbrn:
Thanks for contribution

Issue #717, opened 2019-06-13
On tree views when searching, entries stay collapsed when child elements match search


gianlucamateo:
Not sure if this is intended behavior, labelled as bug as to me it was unexpected.

**Describe the bug**
When the child rows of a parent row in a tree view contain matches to a search item, the parent row is shown, but it is not expanded. The matching child row is therefore invisible until explicit expanding of the parent row.

**To Reproduce**
1. Create a tree view with enabled search, collapse top level rows
2. Search for a term present in a child item
3. See error

**Expected behavior**
I expect search to make all matching items visible, including child items in tree views

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Google Chrome
 - Version: 75.0

gianlucamateo:
As a further bit of info, this Problem does not seem to exist on v1.28.2

gianlucamateo:
I have started work on this

PR #756, opened 2019-06-25
fixed tree sort behavior

gianlucamateo:
## Related Issue
#717

## Description
Reworked the logic for searching in tree views. With these changes, the following rows are shown:

- all items that match the search, regardless of their level in the tree
- all their parent items, regardless of whether they match the search
- all child items of matched items, in a collapsed state

I consider this an intuitive way to search in a tree. Either a leaf node is the target, or some branch node where I want to be able to explore the deeper structure

## Additional Notes
the data-manager is getting awfully large. It might benefit from splitting up into search-manager, filter-manager, etc.


gianlucamateo:
fixed eslint issues now

gianlucamateo:
noticed that column filters have different behavior. Fixed

mbrn:
Hi @gianlucamateo ,

Thanks for contribution. I aggre with you about dataManager. It requires a refactor.

Issue #116, opened 2018-12-13
 Please add functionality to make multi sort possible


SilviaJoycePHMS:
What is the current behavior?
I can only sort for one column. After I click on a second one, it clears the one before.

What is the use-case or motivation for changing an existing behavior?
That I can use a multiple sort based on year followed by month as well.

Eli-Goldberg:
For us this as well this is super important.
I hope this gets implemented.

@mbrn Thanks all for the hard work, you are all great !

mbrn:
Hi @Eli-Goldberg ,

I will apply many more features on table. But nowadays i am very busy and i can reply help-request issues in my limited time. Thanks for your patience

Michael-Roussel:
+ 1 Multi-sort is also important for us. Great work.

priemd:
@mbrn I needed this feature for a project and have things working great.  Wanted to get your take as there are a few ways to go with this.  Currently I added a new property called 'sortObject' to data manager which is passed onto table header through table as a prop.  It is just a simple object { myFieldA: 'asc', myFieldB: 'desc', .... } .  This really takes the place of orderBy and orderDirection, tracking all column sort state.  Wasn't sure if you prefer I just replace orderBy and orderDirection with the new 'sortObject' prop as it would handle both single and multi-sort, but might be a breaking change, need to dig a bit further.  Also, instead of a table option of 'multiSort' I've just used SHIFT key to track whether user wants current sort to be multi, if event.shiftKey is false it is single sort as usual, even if multi-sort had been active for numerous columns and user performs a new sort without holding SHIFT key then it reverts back to single sort.  Any thoughts before I take a pull?

rousselm4:
+1 for multi sort on shift click only, I would also enable multi sort by default

PR #759, opened 2019-06-25
add multi column sort feature

priemd:
## Related Issue
#116 

## Description
Adds column multi-sort feature.

## Additional Notes
- multi-sort is available by default as long as sorting is enabled for table/columns, just hold SHIFT key clicking on header and multi-sort is active.  
- Don't hold SHIFT key when clicking header and it defaults back to single sort.
- There is a new prop for column, 'defaultSortOrder', give it a number which allows default multi-sort.
- Also for onQueryChange a 'sortOrder' property was added which gives the current sort order and for onOrderChange there is an additional argument which is the current sort order.

This could have been a bit cleaner as orderBy and orderDirection aren't necessarily needed anymore as they are both tracked in sortOrder but I left them as is as this would have meant breaking changes for onQueryChange and onOrderChange.

One last note.... I had a 'skip' orderDirection in place where sorting was ignored/reset for the column after clicking header which was previously sorted 'desc', but this would have been a breaking change as well for orderBy and orderDirection I would assume.  Maybe this could be a table option, to opt into 'skip' sorting along with 'asc', 'desc'.

If this works for you I'd be happy to update the docs and throw in an example.  I did add a new table to the demo in the project rather than mucking up with the other tables/data.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -87,7 +87,31 @@ class App extends Component {
       { title: 'Id', field: 'id' },
       { title: 'First Name', field: 'first_name', defaultFilter: 'De' },
       { title: 'Last Name', field: 'last_name' },
-    ]
+    ],
+    multiSort: {
+      columns: [
+        { title: 'strings', field: 'strings', defaultSort: 'asc', defaultSortOrder: 0 },
+        { title: 'letters', field: 'letters', defaultSort: 'desc', defaultSortOrder: 1 },
+        { title: 'bools', field: 'bools', defaultSort: 'asc', defaultSortOrder: 2, type: 'boolean' },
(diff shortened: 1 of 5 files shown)


mbrn:
Hi @priemd ,

Firstly thanks for your contribution. Currently i want to make table more stable instead of adding feature that affect many component. So i don't want to merge this for now. But i will apply multisort later.

1443658:
Hi..
Is this feature applied now?

priemd:
Negative, maintainer rejected because he said he was going to implement
this on his own, maybe 3 months ago.....still no multi column sort.  Been
using it for months, all good so not sure the issues, guess he believes he
can do it better.

hrshtpnt:
Hi, is this feature applied now ? Will this feature be added if not ?@mbrn Can you suggest me an alternative way to achieve this in Material Table.

Issue #615, opened 2019-05-23
Dependent columns


tudro:
Hello,

First thank you for your great work :)

I'm interested if it's possible to change the value of a column based on another one. An example: let's say i have 2 columns, A and B, both dropdowns. I would like that the options available in column B change depending on what i select in column A. 

I tried using the `editComponent` and `props.rowData`, but this rowData doesn't seem to update according to the changes. The values in it are the same from the moment you enter edit mode. So if i change the value in column A, it doesn't update the rowData.

Thank you

tudro:
I've figured it out. In case anybody is interested: in the `onChange` of my custom select defined for the `editComponent` of column A, i check the selected value and then update the state accordingly for the `lookup` of column B. Maybe no the best solution, but it works for my case :)

tudro:
Hi, sorry, i'm still having problems with this for some cases. For example, if i have column A as dropdown and column B as a text field. Depending on what i select in column A, i would like to automatically fill in column B. Is this possible?

Thank you

mbrn:
if you use editComponent for both A and B you can. You will call onChange of B when value of A changed.

tudro:
Thank you for the help, but i can't get it right. How can i access the `props` of B in the `onChange` of A. I mean, when A changes, i have to call the `onChange` of B, but with B's `props`, otherwise i can't update the B field, right?

Thank you

mbrn:
It tried set props of B to component (this.propsB). But when i call double onChange in for A one of them is not working properly. 

So i added onRowDataChange to props of editComponent. This is usage. You can use it for next release. I will publish it tomorrow. 

```javascript
 columns: [
      {
        title: 'Adı', field: 'name', editComponent: props => {
          return (
            <input
              value={props.value}
              onChange={e => {
                var data = { ...props.rowData };
                data.name = e.target.value;
                data.surname = e.target.value.toLocaleUpperCase();
                props.onRowDataChange(data);
              }}
            />
          )
        }
      },
      {
        title: 'Soyadı', field: 'surname', editComponent: props => {
          this.inputBProps = props;
          return (
            <input
              value={props.value}
              onChange={e => props.onChange(e.target.value)}
            />
          )
        }
      },
```

mbrn:
This is done with version 1.39.0

BVH77:
After upgrading to version 1.39.0 when start row edit, in browser console it generate warning: 
```
[...]

PR #770, opened 2019-06-27
Add to prop filter

SmeeK153:
Since onRowDataChange is not being removed as a prop being passed into child components, and it isn't used by them, this is eventually being put onto a simple div element which causes an error because it isn't recognized. By removing this property from the props that are passed down, this error goes away because ultimately the prop function 'onRowDataChange' isn't being loaded onto the subsequent div element at the bottom of the component chain. When a row is updated, the hook 'onRowDataChange' is still working as expected.

## Related Issue
#615 

## Description
To stop the prop function 'onRowDataChange' from being passed down the component chain and outside of this library where it eventually gets placed onto a <div> element and causes an error because it isn't recognized.

## Related PRs
None

* m-table-edit-field.js
* m-table-edit-row.js

## Additional Notes
None

### DIFF
--- a/src/components/m-table-edit-field.js
+++ b/src/components/m-table-edit-field.js
@@ -6,7 +6,7 @@ import PropTypes from 'prop-types';
 
 class MTableEditField extends React.Component {
   getProps() {
-    const { columnDef, rowData, ...props } = this.props;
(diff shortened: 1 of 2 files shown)


markmelvin:
I like this component, but these leaked props are a total pain. Is there any way this merge request can be reviewed and accepted?

Issue #626, opened 2019-05-28
Disable draggable column header


v1nns:
Hi @mbrn, I looked the documentation to disable drag'n'drop column header and only found the flag **column.grouping: false**, but had no effect because I'm not grouping any data. How do I disable the draggable feature in column header? I don't wanna permit to change the column order, otherwise it messes with my workaround to freeze a column.

mbrn:
I can add a columnReordering flag to options for next release.

v1nns:
Hey @mbrn, sup? Did you take a look at this?

mbrn:
Hi @v1nns 

I am not at work and can not work at village. This is Feast of Ramadan days for muslims:). Thanks for patience

v1nns:
Ooops, sorry @mbrn, I had no idea about that. I wish a happy Ramadan Feast to you.

v1nns:
Up

hrafaelveloso:
I think the  #753 solution is what we want for disable the drag header functionality. Can you please merge that @mbrn ? That's a simple opt-out that people could use :)

hrafaelveloso:
Hi @mbrn , I think the #779  solution is what we are looking. That's a simple opt-out and I'm looking forward for that feature.

coltanium13:
hi @mbrn , is there a way to prevent one column from being re ordered when dragging and dropping the headers? I want my first column to always stay in the first position, but the rest of the columns can be moved around. I did not see a column flag property to prevent it from being dragged. Thank you!

jason-zeiler-bayer:
looking for this option as well

PR #776, opened 2019-07-01
Added props to disable draggable header

v1nns:
## Related Issue
#626

## Description
This enables the option to disable the drag-and-drop feature in Table Header.


mbrn:
Hi @v1nns 

I think this is duplicate with #779. What do you think?.  I will close this PR for now.

PR #809, opened 2019-07-09
Feature

theodore-koch:
- Ability to hide the delete and edit icons

## Description
I had a need to hide the delete and edit icons when using the tree view because I am using the first level in the tree to group child records

src/material-table.js


mbrn:
Hi @theodore-koch 

Instead of isEditableHidden, we can add a editButtonProps. What do you think?

theodore-koch:
Yes that would work, also more elegant.

mbrn:
I am closing this PR. We can add editButtonProps instead of this. Can you open an issue about this please?

PR #835, opened 2019-07-16
Enhancement: Use path imports for @material-ui/core

Xammie:
## Description
The [material ui docs state](https://material-ui.com/guides/minimizing-bundle-size/) that in order to minimize bundle sizes you can use full path instead of named imports. For production builds named imports will most likely not be a big issue as webpack's tree shaking solves this. For development builds tree shaking is not enabled and causes the whole `@material-ui/core` module to be imported which could cause slower build speeds and larger bundle sizes. 

This pull request fixes that by replacing the following
```js
import { Button, TextField } from '@material-ui/core';
```
with
```js
import Button from '@material-ui/core/Button';
import TextField from '@material-ui/core/TextField';
```

I was not happy with the build times I was getting on my development builds and tried to optimize all my imports and noticed material-table was not using path imports and that's why I created this pull request

## Additional Notes

This is my first pull request ever :). Feel free to give me feedback on anything about this pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist
 node_modules
 package-lock.json
 yarn.lock
+.idea
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -1,7 +1,9 @@
 /* eslint-disable no-unused-vars */
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import { Icon, IconButton, Tooltip } from '@material-ui/core';
+import Icon from '@material-ui/core/Icon';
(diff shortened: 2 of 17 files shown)


oliviertassinari:
> I was not happy with the build times I was getting on my development builds and tried to optimize all my imports and noticed material-table was not using path imports and that's why I created this pull request

@Xammie Interesting concern! Could you raise your point of view on https://github.com/mui-org/material-ui/issues/16803? We would appreciate it :). Like, what was the order of magnitude of slowness? What's the % win after the change?. Have you considered using a Babel plugin?

Issue #804, opened 2019-07-07
layout breaks as soon as material-table is imported onto the page, 


jayarjo:
As soon as I import `material-table` component onto the page, layout breaks. I suspect that it has something to do with the order in which core material-ui and material-table styles are loaded. Do we have any control over this?

jayarjo:
Happens because `.MuiButtonBase-root` class with all its styles somehow makes to the top. From what I see normally it sits lower in the hierarchy.

jayarjo:
I think this is why it happens: 
![Screenshot from 2019-07-07 12-58-06](https://user-images.githubusercontent.com/150075/60766181-e7811800-a0b6-11e9-8433-17682693d751.png)

Notice all the `Mui*` styles after `makeStyles`. They are basically added twice (there are some of them already defined before `makeStyles`) and I suspect effectively override all the `makeStyles`.

jayarjo:
For some reason I do not observe such behavior in the isolated sandbox: https://codesandbox.io/s/material-demo-2p5qg. So I guess it has to be some setting somewhere. Any ideas?

masagameplay:
This issue is coming from an old Material-UI version in this project (https://github.com/mbrn/material-table/issues/773)

jayarjo:
So what's the fix? Fork it, update and use that one?

kaienbcjdoapwkfldue:
@jayarjo After struggling with this for a while, I now simply updated all the dependecies of my client to their latest version (`npm update`), which solved the problem. So maybe try to update everything related to `material-ui` as well as `material-table`.

Renaud009:
Updating all dependencies didn't actually fixed the issue for me.

alex-fivetalent:
I found with material-table 1.36 to 1.40 fixed the issue, maybe because we had been using hooks with other material ui components and update changed the material-table version of ui/core to 4.

oze4:
This happened to me out of nowhere... I didn't apply any updates, or even install any new packages... literally watched this happen to me from one 'hot reload' to the next. Nothing would fix this for me... I installed `@material-ui/styles` - which made matters worse.

I did not have `@material-ui/styles` installed before all of this started happening (I was hoping it would fix it)..

Just as magically as this issue appeared, it disappeared after uninstalling `@material-ui/styles` - even though I did not have it installed prior to this issue...

Any idea why this happens? This has made me think twice about using this library..

jayarjo:
[...]

Issue #801, opened 2019-07-05
All styling for other components gets disabled when rendering material-table?


sandeshakya:
![iwa-emp-list-4](https://user-images.githubusercontent.com/35713511/60737271-64758b80-9f17-11e9-813e-849edc068219.gif)

I didn't have this issue when testing out the library on a test project.

My dependencies are:"@material-ui/core": "^4.1.3",
    "@material-ui/icons": "^4.2.1",
    "@material/menu": "^3.0.0",
    "aws-amplify": "^1.1.29",
    "aws-amplify-react": "^2.3.9",
    "lodash": "^4.17.11",
    "material-table": "^1.40.1",
    "material-ui-popup-state": "^1.4.0",
    "moment": "^2.24.0",
    "notistack": "^0.8.8",
    "react": "^16.8.6",
    "react-block-ui": "^1.1.4",
    "react-data-grid": "^6.1.0",
    "react-dom": "^16.8.6",
    "react-material-ui-form-validator": "^2.0.9",
    "react-router-dom": "^5.0.1",
    "react-scripts": "3.0.1"

I would appreciate some help on what is causing this.

Thank you!

sandeshakya:
found answer here
https://github.com/mbrn/material-table/issues/773

Issue #773, opened 2019-06-28
loss of style with v4 of the wm material-ui


Annderson:
all the pages I'm using the material-table. Loses the applied style, mainly those that are in the theme

"dependencies": {
    "@material-ui/core": "^4.1.3",
    "@material-ui/icons": "^4.2.1",
    "@material-ui/styles": "^4.0.0-alpha.0",
    "axios": "^0.18.0",
    "babel-polyfill": "^6.26.0",
    "camelcase-keys": "^5.0.0",
    "classnames": "^2.2.6",
    "connected-react-router": "^6.3.2",
    "history": "^4.9.0",
    "leaflet": "^1.5.1",
    "lodash": "^4.17.11",
    "material-table": "1.37.0",
    "moment": "^2.24.0",
    "query-string": "^6.5.0",
    "react": "^16.8.6",
    "react-dom": "^16.8.6",
    "react-leaflet": "^2.3.0",
    "react-notification-system-redux": "^2.0.0",
    "react-redux": "^6.0.1",
    "react-router": "^5.0.0",
    "react-router-dom": "^5.0.0",
    "react-scripts": "2.1.8",
    "recharts": "^1.5.0",
    "recompose": "0.30.0",
    "redux": "^4.0.1",
    "redux-devtools-extension": "^2.13.8",
    "redux-persist": "^5.10.0",
    "redux-saga": "^1.0.2",
    "snakecase-keys": "^2.1.0",
    "typeface-roboto": "^0.0.54",
    "urijs": "^1.19.1",
    "validator": "^10.11.0"
  },

json2d:
same
```
    "@material-ui/core": "^4.1.3",
    "material-table": "^1.36.4",
```

any hints on better working versions?

masagameplay:
The issue is probably coming from an old Material UI version in the package.json (https://github.com/mbrn/material-table/blob/master/package.json#L75). It seems like the other packages are very outdated too.

json2d:
`material-table@v1.36.6` is the last version before they upgraded to `@material-ui/core@v4`, so downgrading to that might work
https://github.com/mbrn/material-table/blob/v1.36.6/package.json#L75

Also shouldnt `@material-ui/core` be a peer dependency?

json2d:
yep, downgrading to `material-table@v1.36.6` seems to have worked for the style loss/bug i was experiencing

Annderson:
Yes, before v4 of material-ui worked, but I need to use v4

[...]

PR #851, opened 2019-07-22
Make @material-ui/core a peer dependency

luzzif:
## Description
Material UI complains when multiple library instances are initialized in a single application.
The issue can be easily solved by making `@material-ui/core` a peer dependency plus development dependency, instead of having it as a production dependency only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,10 +57,12 @@ If you use an older version of react we suggest to upgrade your dependencies or 
 To install material-table with `npm`:
 
     npm install material-table --save
+    npm install @material-ui/core --save
 
 To install material-table with `yarn`:
 
     yarn add material-table
+    yarn add @material-ui/core
 
 #### 2.Add material icons
 
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "@babel/plugin-transform-runtime": "7.1.0",
     "@babel/preset-env": "^7.2.0",
     "@babel/preset-react": "7.0.0",
+    "@material-ui/core": "^4.0.1",
     "babel-eslint": "10.0.1",
     "babel-loader": "^8.0.4",
     "babel-polyfill": "^6.26.0",
@@ -72,7 +73,6 @@
   },
   "dependencies": {
     "@date-io/date-fns": "^1.1.0",
-    "@material-ui/core": "^4.0.1",
     "@material-ui/pickers": "^3.0.0",
     "classnames": "^2.2.6",
     "date-fns": "^2.0.0-alpha.27",
@@ -81,5 +81,8 @@
     "prop-types": "^15.6.2",
     "react-beautiful-dnd": "11.0.3",
     "react-double-scrollbar": "0.0.15"
+  },
+  "peerDependencies": {
+    "@material-ui/core": "^4.0.1"
   }
 }


luzzif:
IMHO this is pretty important, have a look [here](https://material-ui.com/getting-started/faq#i-have-several-instances-of-styles-on-the-page).

It's a pretty trivial PR too, I would love for this to get approved fast.

nareshbhatia:
It appears that you have only moved material-ui/core to peerDependencies. You should also specify it in devDependencies. This will have the advantage that you can easily do local development in the repository, while not including the bundle in the build.

Also I would like to see zero `dependencies`. All dependencies should be moved to `devDependencies` and `peerDependencies`.

I have the same use case with my [React Force](https://github.com/nareshbhatia/react-force) library that depends on Material UI. See this [package.json](https://github.com/nareshbhatia/react-force/blob/master/package.json) as an example.

luzzif:
Well I have moved `@material-ui/core` in `devDependencies` right from the start, so I don't know what you are talking about.

As far as making all dependencies peers, I moved `@material-ui/core` just because I know that multiple instances in the same application don't play too well. 
IMHO there are some dependencies that make sense as prod dependencies and not peers.

nareshbhatia:
@luzzif, good that `@material-ui/core` is already in `devDependencies` - it's my miss on my part.

As for other dependencies, I feel that only those dependencies should be included that no one else is likely to include in their own apps. For example, I already have `@material-ui/pickers` & `classnames` in my app.

luzzif:
Yeah, it's definitely debatable, but personally in my app I don't directly use `classnames`, so I wouldn't like my `package.json` to get cluttered by it.

I would love to see in my `package.json` only those dependencies which I directly use or which I directly need (the only exceptions being dependencies which create problems when multiple instances are found in the same app: those should always be peers).

P.S.: sorry if I sounded jerky when I told you that I already had added `@material-ui/core` to the `devDependencies`, I obviously didn't mean to.

nareshbhatia:
As a side note, it might be good to use storybook to develop & demo the various features. It eliminates the need for creating CodeSandboxes and/or separate apps for each feature. Just a thought.

nareshbhatia:
nw, @luzzif - no offense taken :-)

I get you point about adding more libraries into peer dependencies. It's a double-edged sword!

Personally, I would like to use material-table only for its sorting feature. I would like to see it come to a point where including it in my app does not add too much weight. Until then I will simply add a poor-man's sort on top of Material UI's table (which adds only 1K gzipped to my app).

luzzif:
Can someone approve this?

victor-accarini:
@mbrn This is pretty important see #773, #801 and #804.

luzzif:
It is important indeed. It baffles me how much time a little PR like this takes to get approved...

victor-accarini:
I believe the author was away from the project last month, it will probably take a while until he get up to date with the PRs. Last time he merged mine in less than a day, he is quite fast 👍

mbrn:
> I believe the author was away from the project last month, it will probably take a while until he get up to date with the PRs. Last time he merged mine in less than a day, he is quite fast 👍

Thank @victor-accarini . For now i will merge PRs only. Then we will work on issues with @emrekara37.

luzzif:
So is this going to be approved? Material Table totally breaks the page's styles when used with the last few versions of Material UI. It's pretty bad...

eps1lon:
For anybody having issues feel free to post your lockfiles here and I should be able to provide you with a patch that fixes those issues.

Issue #745, opened 2019-06-21
Remove warning for invalid column type


sornman:
**Is your feature request related to a problem? Please describe.**
When specifying a custom column type, the following error occurs:

`Warning: Failed prop type: Invalid prop 'columns[1].type' of value 'textarea' supplied to 'MaterialTable', expected one of ["string","boolean","numeric","date","datetime","time","currency"].`

**Describe the solution you'd like**
Remove the warning if it encounters a type that it doesn't recognize.

**Additional context**
I am using custom column types, specifically 'select', to trigger a react-select dropdown, and 'textarea'. To accomplish this, I use:

`components={{
EditField: props => {
if (props.columnDef.type === 'select') { ...
} else if (props.columnDef.type === 'textarea') { ...
}
}
}}`

Everything seems to be working great, except for the above warning message in the console.

Domino987:
IS this still a problem?

PR #858, opened 2019-07-24
Fixes issue #745

emrekara37:
## Related Issue
#745 

## Description
Implemented fix as described in the issue


mbrn:
Hi @emrekara37 

It does not seem a good solution. Developer should not use type field for a different purpose. We may try to find another solution for this problem.

PR #859, opened 2019-07-24
Performance improvements

emrekara37:
## Description
`React.Component` to `React.PureComponent`


mbrn:
Hi @emrekara37 

We don't need improvements for now. We should focus to bugs firstly. Then we can make better improvements. Especially on algorithms.

Issue #860, opened 2019-07-24
The search feature should wait for user to stop typing


nvtin:
When I implement the search box with remote data. It is always jerky when I type the text. Because It always make the update the search text and do a query to get data when I type.

Solution to fix this issue is make the timeout to wait for user to stop typing at least 500 milliseconds. So the search will be called after user stop typing 500 ms.

Actually we will have options to user set the time if they want to custom:
```
options: {
  ..
  search: true,
  searchProps: {
    waitForTyping: 500
  },
 ..
}
```

oze4:
Check out `debounceInterval` under `Options` [in the docs](https://material-table.com/#/docs/all-props) - the default is `200ms`.

Should be able to do something like:

```
<MaterialTable
    ...
    options={{
        ...
        debounceInterval: 500,
        ...
    }}
    ...
/>
```

nvtin:
@oze4 Thanks so much. I read the docs but I missed it.

PR #861, opened 2019-07-24
add the wait time for typing in search box

nvtin:
## Related Issue
Relate the Github issue with this PR using `#860`

## Description
This PR is to fix the issue for search: `#860 The search feature should wait for user to stop typing`

When I implement the search box with remote data. It is always jerky when I type the text. Because It always make the update the search text and do a query to get data when I type.

Solution to fix this issue is make the timeout to wait for user to stop typing at least 500 milliseconds. So the search will be called after user stop typing 500 ms.

Actually we will have options to user set the time if they want to custom:

options: {
  ..
  search: true,
  searchProps: {
    waitForTyping: 500
  },
 ..
}

### DIFF
--- a/src/default-props.js
+++ b/src/default-props.js
@@ -85,6 +85,9 @@ export const defaultProps = {
     showFirstLastPageButtons: true,
     showSelectAllCheckbox: true,
     search: true,
+    searchProps: {
+      waitForTyping: 500
+    },
     showTitle: true,
     showTextRowsSelected: true,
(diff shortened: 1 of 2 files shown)


nvtin:
@mbrn Can you take the time to review it?
It is for this issue: https://github.com/mbrn/material-table/issues/860

nvtin:
I will close this PR because we have an option to take care of it already
debounceInterval: debounce interval for search and filter

Issue #618, opened 2019-05-24
How to specify DetailPanel column width?


behrouz-s:
Detail Panel's icon is shown in vertical alignment when number of column is more than 8.
![download](https://user-images.githubusercontent.com/1307397/58333792-e8335700-7e46-11e9-8e60-95ed8ec42f27.png)
![download (1)](https://user-images.githubusercontent.com/1307397/58334037-6132ae80-7e47-11e9-9ec9-0ec63d2b4fc9.png)

mbrn:
This should be done with version 1.36.6

behrouz-s:
The problem was fixed in mentioned version, but I have same problem in version 1.39.0!

mbrn:
Let me take a look again.

behrouz-s:
Would you please fix the problem, it could be seen in your documentation too.
![download (1)](https://user-images.githubusercontent.com/1307397/61156723-c0827600-a4fc-11e9-9c49-f55461a9bd13.png)

PR #868, opened 2019-07-24
Fixes issue #618

emrekara37:
## Related Issue
#618 

## Description
Fix detail panel column width 

Screenshot_1

### DIFF
--- a/src/components/m-table-body-row.js
+++ b/src/components/m-table-body-row.js
@@ -96,7 +96,7 @@ export default class MTableBodyRow extends React.Component {
     else {
       return (
         <TableCell size={this.getElementSize()} padding="none" key="key-detail-panel-column">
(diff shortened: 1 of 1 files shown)


jurcaua:
I believe this should instead be a change of:
`<div style={{ width: 42 * this.props.detailPanel.length, textAlign: 'center', display: 'inline-block' }}>`
to
`<div style={{ width: 42 * this.props.detailPanel.length, textAlign: 'center', display: 'flex' }}>`

emrekara37:
hi @jurcaua  
i uptaded the way you said
thank you

jurcaua:
Hey @emrekara37, the width change may not be needed, but I'm not sure for all use-cases. 
For my needs the change to "flex" was enough.

@mbrn can we get this small fix in soon if possible?

Issue #466, opened 2019-04-14
Don't use custom render for grouping


revskill10:
Currently, group dragging doesn't work if a column has a custom render function.

## Solution

Use a `dragRender` function which specify the string to show when grouping

victor-accarini:
@mbrn I have the same problem and would like to help fix this...

Do you prefer a specific function to handle group rendering or should we handle everything inside the `render` function?

For the latter option it would be nice to send a third argument with all group rows data, so we could choose which rows to use when rendering for the group...

csforbes:
For anyone else dealing with this issue, I was able to get past this issue by writing a "customRender" utility function.

```
function customRender(value, renderType, renderFunc, field, ...args) {
    if (renderType === 'row') {
        return renderFunc(value[field], ...args);
    }
    if (renderType === 'group') {
        return renderFunc(value, ...args);
    }
}

//normal render function
function renderCellData(status) {
   //do stuff
}

//using in column object
...
field: 'status',
render: (value, renderType) => customRender(value, renderType, renderCellData, 'status')
```

KLM-GIT:
Thanks @csforbes - your solution works great for grouping by a rendered column value.

somandubey:
> For anyone else dealing with this issue, I was able to get past this issue by writing a "customRender" utility function.
> 
> ```
> function customRender(value, renderType, renderFunc, field, ...args) {
>     if (renderType === 'row') {
>         return renderFunc(value[field], ...args);
>     }
>     if (renderType === 'group') {
>         return renderFunc(value, ...args);
>     }
> }
> 
> //normal render function
> function renderCellData(status) {
>    //do stuff
> }
> 
> //using in column object
> ...
> field: 'status',
> render: (value, renderType) => customRender(value, renderType, renderCellData, 'status')
> ```

Thanks @csforbes, this was a nice hack to solve the same!, appreciate the share.

thalysonalexr:
Hello Guys!
This solution don't work for  me!

I have a Boolean column and I want the label "Sim" to true and "Não" to false, but when doing:

```ts
render: (rowData, type) => {
        if (type === 'row') return rowData.hasPages ? 'Sim' : 'Não'
        return !rowData ? 'Não' : 'Sim'
},
```

works for row but when I group I get two values ​​true, and does not render any value for label "No". Has anyone had this same problem?

[...]

PR #872, opened 2019-07-25
Change column rendering structure for grouped columns

victor-accarini:
## Related Issue
#466 

## Description
When grouping values the render function, provided by the column
definition, no longer receives the whole data from the row element just
the value used to group rows.

This behavior was troublesome since some columns used different values
from the row to properly render its information, therefore it would be
impossible to render based on just the grouping value.

Now the group rendering was moved to a specific rendering function named
renderGroup which receives the grouped value and all rows from that
specific group.

* TableCell
* TableGroupRow

## Additional Notes
This introduces an API change for the render function, so it might break existing applications

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -126,6 +126,13 @@ class App extends Component {
                       src={rowData.avatar}
                     />
                   ),
+                  renderGroup: (avatar, groupData) => (
+                    <img
+                      style={{ height: 36, borderRadius: '50%', bottom: -10, position: 'relative' }}
+                      src={avatar}
+                    />
+
+                  )
                 },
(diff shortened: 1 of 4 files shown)


mbrn:
Hi @victor-accarini,

We can implement something like this at version 2. Because it contains a breaking change.

hamdi-islam:
Hello @mbrn,
when's version 2 coming out ?

Clem-:
Hi,
Is there a way to fix this behavior and get the whole dataset in the render function?

Thanks

victor-accarini:
@mbrn Any news of version 2?

Issue #655, opened 2019-06-01
Tree Data get Selected Rows


KyriacosP:
![material-tebleTreeData](https://user-images.githubusercontent.com/44581807/58754472-398fb600-84d9-11e9-82ee-068b8e213527.PNG)

onSelectionChange seems to be working wrong here the rows argument contains multiple copies of each row selected although the toolbar count seems to be working fine

mbrn:
Hi @KyriacosP 

Thanks for issue. This is a bug. I will resolve this problem asap.

waisbek:
Hello,

I have the same issue. @mbrn do you have a deadline to fixed this bug, please?

Thanks!

KyriacosP:
i removed the duplicates from the selected array based on their id it works well in my case but it may be slow if your selection array is too large. here is my code:
`unique(array) {
    let arr = [];
    for(let i = 0; i < array.length; i++) {
      if(!arr.some(x=>x.id===array[i].id)) {
        arr.push(array[i]);
      }
    }
    return arr;
  }`

cristiadis:
@emrekara37 This is not fixed. I still get duplicate values when selecting, in version 1.46.0

charlesc32:
I am also still seeing this issue. I think the fix only addressed if parent row duplicates. Children are still being duplicated.

estubmo:
I'm also experiencing this issue.
![image](https://user-images.githubusercontent.com/5949794/85954281-8d87fb00-b976-11ea-9a31-f75282251392.png)

Issue #545, opened 2019-05-07
CheckAll box not working probably


allennyzhang:
**Describe the bug**
if any records' was selected, then will dynamic display the button `Schedule`, but the `CheckAll` checkbox doesn't selected correctly.

![image](https://user-images.githubusercontent.com/24515577/57306787-03e3f100-7116-11e9-9306-202dd26f202d.png)
![image](https://user-images.githubusercontent.com/24515577/57306818-11997680-7116-11e9-9863-2a04aa1a6b2e.png)
![image](https://user-images.githubusercontent.com/24515577/57306840-1b22de80-7116-11e9-997d-a1ffb3ad49aa.png)
![image](https://user-images.githubusercontent.com/24515577/57306868-237b1980-7116-11e9-963a-409a867870d3.png)

mbrn:
Hi @allennyzhang ,

Which version of MT do you use? Because i tested it with latest version and it works.

emrekara37:
I am closing this issue. You can reopen if you need.

Issue #498, opened 2019-04-22
Nested expanded rows collapses on selection


ozandogrultan:
**Describe the bug**
Related issue #443 . Still not fixed for nested rows that have depth more than 1.

**To Reproduce**
1. Render a table that is nested more than 1 level
2. Select any nested child or parent row

**Expected behavior**
Row collapse/expand not triggered by selection.

**Screenshots**
![Apr-22-2019 18-58-15](https://user-images.githubusercontent.com/5598573/56509824-ac982b00-6530-11e9-9e41-343843f1a17c.gif)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version 1.32.0

mbrn:
Please upgrade to 1.34.0

GienekGenin:
@mbrn Still not fixed for nested rows that have depth more than 1.

![BitStreamGoogleChrome20190](https://user-images.githubusercontent.com/27743401/57362073-450feb80-717e-11e9-8a25-00afd604182e.gif)

Desktop:
OS: Windows 10
Browser: Chrome
Version `"material-table": "^1.36.1"`

mbrn:
Hi @GienekGenin,

I will look at this asap.

theodore-koch:
Was this ever resolved?

benjaminvdb:
![select_collapse](https://user-images.githubusercontent.com/8875786/62302755-2d918780-b47b-11e9-8a30-4de02f1a1e2f.gif)

I'm on version 1.41.0 and I'm having the same issue: the tree collapses whenever I select a row.

v1nns:
Up

raph90:
Yikes we're all in the same boat... :(

emrekara37:
Hi , i am working on it 
We will fixed it asap.

ruthwe:
any news?

lsansaa:
Not sure if is related, but this happens with actions buttons if do something like showing a modal/dialog.

Is there a way to get the on click event of the expand button? Got an idea of how quick fix this, saving on a state the isTreeExpanded attribute of each row and then using it to render the row. The onTreeExpandChange event won't do the thing since it gets trigger after the row is rendered, not after clicking the button (Already try it with this one)

v1nns:
Any update about it?

PR #875, opened 2019-07-25
fix tree data selected rows #655

emrekara37:
## Related Issue
#655
#545 
#498 
## Description
fix tree data selected row count 
Screenshot_1


mbrn:
Hi @emrekara37 ,

Are you sure that is solution of related issues?

emrekara37:
> Hi @emrekara37 ,
> 
> Are you sure that is solution of related issues?

in any case count increase when click select button on the parent rows 
```js  
this.selectedCount = this.selectedCount + (checked ? 1 : -1);
```
therefore count is wrong

PR #887, opened 2019-07-26
feat: export unstyled components

spurreiter:
The following components are only available as being styled. This prevents further inheritance.
This PR exports the following components as "unstyled" variants

- MTableHeaderUnstyled
- MTableSteppedPaginationUnstyled
- MTablePaginationUnstyled
- MTableToolbarUnstyled

## Related Issue

None

## Description

Export all components also as an unstyled variant to allow custom extension through inheritance.

## Related PRs

None

* None - available components are exported-

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -1,31 +1,14 @@
-import MTableAction from './m-table-action';
-import MTableActions from './m-table-actions';
-import MTableBody from './m-table-body';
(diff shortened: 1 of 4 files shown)


mbrn:
Hi @spurreiter 

Thanks for contribution. But i will make all sub component styled on second version of MT with a proper way.

Issue #900, opened 2019-07-31
Setting detailPanelColumnAlignment to right misaligns row while editing


Jim-Gambit:
**Describe the bug**
First of all great work! Material Table has made my work a lot easier. I am trying to create a table with multiple detail panels. I want the icon column of the panels on the right hand side and I am using detailPanelColumnAlignment props with value 'right' in option. This works fine but when I try to edit a row, column misaligns with the header i.e. it shifts right by one column.

**To Reproduce**
1. Make the table rows editable
2. Create a multiple details panel
3. In options, set detailPanelColumnAlignment: 'right'
4. Try editing any row

**Expected behavior**
Rows should not shift or misalign from headers while editing even if detailPanel column is right align.

**Desktop (please complete the following information):**
 - OS:  Ubuntu 16.04
 - Browser: Chrome
 - Version: 75.0.3770.142

mbrn:
Hi @Jim-Gambit ,

We will resolve this bug asap.

PR #902, opened 2019-07-31
Fixes issue #900

emrekara37:
## Related Issue
 #900

### DIFF
--- a/src/components/m-table-edit-row.js
+++ b/src/components/m-table-edit-row.js
@@ -169,7 +169,7 @@ export default class MTableEditRow extends React.Component {
     }
 
     // Lastly we add detail panel icon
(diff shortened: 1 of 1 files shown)


mbrn:
Hi @emrekara37 ,

What happens if detailPanelColumnAlignment is 'right'?

In addition, can you check filterRow for detailPanelColumnAlignment left/right cases?

emrekara37:
> Hi @emrekara37 ,
> 
> What happens if detailPanelColumnAlignment is 'right'?
> 
> In addition, can you check filterRow for detailPanelColumnAlignment left/right cases?

I updated the way you said and tested,
if `detailPanelColumnAlignment` is `right` adding to the end of the list otherwise  adding to the start of the list

Issue #906, opened 2019-07-31
Maintain state on expanded tree row


theodore-koch:
When the material-table renders on data change I want the same expanded rows to still be expanded. Is there a state the material-table exposes to support this?

theodore-koch:
Added new PR feature - PR #912

PR #912, opened 2019-08-01
feat: expanded rows option to control expand state

theodore-koch:
## Related Issue
#906 

## Description
Ability to control which rows are expanded using the row id

* default_props.js
* material-table.js
* prop-types.js
* data-manager.js


mbrn:
Hi @theodore-koch ,

I solved this problem with an eaiser way. You use row.tableData.isTreeExpanded: true . 

Check latest this commit please:

https://github.com/mbrn/material-table/commit/0e632cc32372106f57d6378ac108200b3fa73b94

Issue #384, opened 2019-03-27
When data is grouped the exported csv only contains the column titles


crotten:
**Describe the bug**
When I try and export grouped data the csv only contains the column titles. If I remove the groups the exported csv has all the expected data.

**To Reproduce**
1. Create a table with groups
2. Drag a group to the top to make it active
3. Export the data
4. Open exported csv file

**Expected behavior**
When exporting grouped data the data would export successfully.

jafstar:
I see this bug when using data in a sub-object such as form.info.location as opposed to just 'location'

jafstar:
In /dist/m-table-toolbar.js I simply check for a '.' and split the string, it renders the export csv as expected. Not sure where this is in source, but will see if I can submit a PR for this.

`var data = _this.props.renderData.map(function (rowData) {
          return columns.map(function (columnDef) {
            let splitField = columnDef.field.split('.')
            return String(columnDef.field).includes('.') ? rowData[splitField[0]][splitField[1]] : rowData[columnDef.field];
            // return columnDef.lookup ? columnDef.lookup[rowData[columnDef.field]] : rowData[columnDef.field];
          });
        });`

karanness86:
Hello @mbrn @emrekara37 i am still having the issue described here even though i updated to the latest version (1.54.2) which has this fix. Does it have any other version dependency? As @crotten described, the export csv has the data when there is no grouping. Thank you for your assistance.

karanness86:
I ended up using following the export example to write a separate exportCsv function. Thats when i figured out that when grouping, the data structure needs to be curated before it can be used to create the csv. After that i just used the code from m-table-toolbar.js to recreate the export function using the curated data. This will work for now but when there is additional grouping then the data needs to be curated even more before calling the csv builder process.

PedrinhowVieira:
My current version is ("material-table": "1.67.1"), and this issue is still occurring, I have a table that has a default grouping, when trying to export it, only the titles are exported, not the content of the table.

sinclas:
1.69.3 and cant export with grouping.  not using a renderer either,  simple text fields.  both CSV and PDF not working.

piotrpodyma10:
yes it's true. SOmebody can solve it?

TatyanaVl:
It seems that this still has not been fixed.

PR #915, opened 2019-08-02
Fix when drag column export csv column order bug #384

emrekara37:
## Related Issue
#384 

## Description
Simple words to describe the overall goals of the pull request's commits.


emrekara37:
> You should add sort function after filter like in render method of MFilterRow

Updated.

PR #927, opened 2019-08-05
Fix row `onClick` typedef

KATT:
## Description

Callback is the row clicked, not the whole set


mbrn:
Hi @KATT 

Thanks for contribution. For free actions it sends array of rowData. So i applied both types.

PR #928, opened 2019-08-05
Feature/conditional should expand

dnn5b:
## Description
An optional property to define which rows should be expanded in a tree table. The property expects a function. If the `defaultExpanded` is already set to true the new `shouldExpand` property is ignored.

branch | PR
------ | ------
 feature/conditional-should-expand | [Link](https://github.com/dnn5b/material-table/tree/feature/conditional-should-expand)

## Additional Notes
This feature helps to expand only a few rows instead of all of them (e.g. specific data type).


gianlucamateo:
I know this is similar to #912 , but @mbrn I really think we should incorporate one of these PRs. Setting the expansion state by setting a variable in the tabledata of each row violates SOLID principles by requiring referencing components to know about the inner workings of material-table. If the table-data approach is ever changed, or the variables renamed for whatever reason, any component relying on this method would break. The added layer of abstraction provided by props would be highly desireable.

mbrn:
Hi @dnn5b ,

It requires a few changes. You should implement this on DataManager.changeTreeExpand method, too. In addition can you change shouldExpand as isTreeExpandable .

dnn5b:
Hi @mbrn,

I'm not sure whether I understand you correctly:
- "_change shouldExpand as isTreeExpandable_": do you mean I should rename the property to `isTreeExpandable`? If no, then please clarify :) If yes I don't think that this is a good name. The property should only specify which rows are collapsed by default. If the user decides to expand them afterwards it should still work. So it's more like `isExpandedByDefault`.
- "_implement on DataManager.changeTreeExpand_": see first point. If the function should only change the default state it can be ignored in the changeTreeExpand method. Or do I miss anything here?

mbrn:
> Hi @mbrn,
> 
> I'm not sure whether I understand you correctly:
> 
> * "_change shouldExpand as isTreeExpandable_": do you mean I should rename the property to `isTreeExpandable`? If no, then please clarify :) If yes I don't think that this is a good name. The property should only specify which rows are collapsed by default. If the user decides to expand them afterwards it should still work. So it's more like `isExpandedByDefault`.
> * "_implement on DataManager.changeTreeExpand_": see first point. If the function should only change the default state it can be ignored in the changeTreeExpand method. Or do I miss anything here?

Hİ @dnn5b ,

I understand you now :). It will only change behavior of default expanded or not. So why you don't implement on isExpandedByDefault. Let it being function or boolean value. Just check type of it and use boolean value or call function. 

I mean this: ```defaultExpanded?: boolean | ((rowData: any) => boolean);```

dnn5b:
@mbrn that's a good idea. I refactored my solution: Now only the option `defaultExpanded` is used as you suggested.

vladvonpr:
@dnn5b Is it possible for you to show me a code snippet of this feature working? 

I saw this in the demo commit you did (which I can't find in the latest code anymore):
` defaultExpanded: row => row.surname === 'C'`

I'm trying things like the following:
`defaultExpanded: row => row.Year === 2020`
or
`defaultExpanded: row => row.DefaultExpanded === true` (where DefaultExpanded is a boolean column in the data)

vladvonpr:
@dnn5b I think my issue might be that this feature only works for Tree and not the standard group.

On the data-manager.js, in the groupData function, it's not using the function alternate. If it's not boolean, then it's false:
```
if (!group) {
            const path = [...(o.path || []), value];
            let oldGroup = this.findGroupByGroupPath(
              this.groupedData,
              path
            ) || {
              isExpanded:
                typeof this.defaultExpanded === "boolean"
                  ? this.defaultExpanded
                  : false,
            };
```

But further down, in the treefyData() function it does use the function alternate if it's not Boolean:
```
if (rowData.tableData.isTreeExpanded === undefined) {
          var isExpanded =
            typeof this.defaultExpanded === "boolean"
              ? this.defaultExpanded
              : this.defaultExpanded(rowData);
          rowData.tableData.isTreeExpanded = isExpanded;
        }
```

Issue #936, opened 2019-08-06
Make sort possible with three states


hrafaelveloso:
Right know we can only sort the column by ASC and DESC. 

I think it lacks the possibility to be able to sort with 3 states: ASC, DESC, NORMAL. In the third click in sort column, the data should be like it was in the first time the table renders.

You can sort the column name in [this codesandbox](https://codesandbox.io/s/confident-grass-n46l7), and it starts with : Mehmet, Zerya and Rafael. The first sorting click makes it into Mehmet, Rafael, Zerya and the second into Zerya, Rafael, Mehmet. It's not possible to have Mehmet, Zerya, Rafael again without reload the page.

This behaviour is seen is spotify where you sort the artist/album name and it has 3 states.

PR #937, opened 2019-08-06
Update set sort direction logic on table header component, to enable …

gredy2608:
…reset sort direction when a column with sort direction 'desc' clicked

## Related Issue
#936 

## Description
Update the logic on m-table header component to enable functionality to reset the sort direction by clicking the column header with sort direction 'desc'. 
The flow for sort direction will be like this: "asc" => "desc" => "" => "asc" => ...

## Related PRs

branch | PR
------ | ------

src/components/m-table-header.js
*

## Additional Notes


mbrn:
Hi @gredy2608 

Thanks for contribution. But it is not working. I think that it requires more changes.

mbrn:
@hrafaelveloso solved this problem with #938

PR #938, opened 2019-08-06
Add third state to sort column

hrafaelveloso:
## Related Issue
Relate the Github issue with this PR using `#936`

## Description
In #936 I wanted to be able to sort with 3 states. In this PR it's now possible.

It's now possible to have three states to sort one column.


adamsoffer:
Correct me if I'm wrong, but it appears as though this is now the default setting which is...weird. I have to click 3 times on a column to sort. It should be two. Related: #985

adamsoffer:
cc @hrafaelveloso @mbrn

PR #939, opened 2019-08-06
ActionFeedback as a component prop

skrhlm:
## Related Issue
- No related issues. 

## Description
- Added a prop to `components` called `ActionFeedback`. This allows users to supply their own Typography component for the prompt that is shown when trying to update or delete a row.
- I fixed the propagation of the localized props regarding the editRow, to avoid these being declared at multiple places, and mismatching.
- Added a dummy onRowDelete to the demo.js to display this functionality.

## Related PRs
- No related PRs

* MTableEditRow

## Additional Notes
As this is a new feature it will probably not break anything.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -89,6 +89,16 @@ class App extends Component {
                           resolve();
                         }, 1000);
                       }),
+                      onRowDelete: id =>
+                      new Promise((resolve, reject) => {
+                        setTimeout(() => {
+                          {
+                            /* 
+                            this.setState({ data }, () => resolve()); */
+                          }
+                          resolve();
+                        }, 1000);
+                      }),
                   }}
                 />
               </Grid>
--- a/src/components/m-table-edit-row.js
+++ b/src/components/m-table-edit-row.js
@@ -1,15 +1,14 @@
 /* eslint-disable no-unused-vars */
 import TableCell from '@material-ui/core/TableCell';
 import TableRow from '@material-ui/core/TableRow';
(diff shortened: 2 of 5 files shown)


mbrn:
Hi @skrhlm ,

I think to change style of delete text we don't need a component overriding. Instead of it we can use a style parameter to pass it. What do you think ?

skrhlm:
Hello @mbrn and thanks for the review! 

I kind of agree with you, however I feel as if this would result in styles being duplicated in many cases. It's often the case that you have a readymade component, which you have to copy the styles from to achieve the same within a component library.
However, I do like the idea of going class-based/style-tags all the way ( like MUI-theme ).

If I've failed to convince you I'll of course convert this PR into a style-based solution instead :)

mbrn:
> Hello @mbrn and thanks for the review!
> 
> I kind of agree with you, however I feel as if this would result in styles being duplicated in many cases. It's often the case that you have a readymade component, which you have to copy the styles from to achieve the same within a component library.
> However, I do like the idea of going class-based/style-tags all the way ( like MUI-theme ).
> 
> If I've failed to convince you I'll of course convert this PR into a style-based solution instead :)

Hi @skrhlm ,

I am planning to implement styled based component on MT. For version 2, we have many plans. But now i want to focus to bugs and resolve the issues.

jayarjo:
I'd say provide a way to override each and every component @mbrn :)

mbrn:
I don't want to make every detailed component overridable for now. I want to close this pr just for now. We may later imply a better solution for problems like this.

Issue #766, opened 2019-06-26
Add support for className


mviens:
**Is your feature request related to a problem? Please describe.**
While there is support for `<MaterialTable style={{}} ... />`, there does not seem to be support for `<MaterialTable className={} ... />`.

**Describe the solution you'd like**
I want to be able to define a CSS class name using `<MaterialTable className={classes.myTable} ... />` or `<MaterialTable className='myTable' ... />`..

**Describe alternatives you've considered**
Wrapping the `<MaterialTable />` in a `<div>` and using CSS to find it that way (hacky).

pak11273:
So we don't bloat the codebase, there is no need to add a className since we can already style material table with either style or by overriding the Container component like this:

````
<MaterialTable
     Container: props => (
      <div style={{ ... }}>
        <div {...props} />
      </div>
    )
</MaterialTable>
````

**Comment by sabbin** [comment]
Thanks for the response, I tried that but that actually wraps the `Toolbar` inside a `div` which has `padding:0`. It does not apply the `padding` prop to the `MToolbar` component.

The result is the following 

```html
<div style="padding: 0px;">
    <div class="class="MuiToolbar-root MuiToolbar-regular MTableToolbar-root-420 MuiToolbar-gutters">...</div>
</div>
```

I need to remove the padding from the child `div`, in this case the `MuiToolbar-root`

pak11273:
Do it inside the Toolbar as shown in the [documentation](https://material-table.com/#/docs/features/component-overriding):

````
components={{
          Toolbar: props => (
            <div>
              <MTableToolbar {...props} />
              <div style={{padding: '0px 10px'}}>
                <Chip label="Chip 1" color="secondary" style={{marginRight: 5}}/>
                <Chip label="Chip 2" color="secondary" style={{marginRight: 5}}/>
                <Chip label="Chip 3" color="secondary" style={{marginRight: 5}}/>
                <Chip label="Chip 4" color="secondary" style={{marginRight: 5}}/>
                <Chip label="Chip 5" color="secondary" style={{marginRight: 5}}/>
              </div>
            </div>
          ),
        }}
````

sabbin:
That does not work either.. it just creates another div beside the `<MTableToolbar/>` 

Look at the images below

![Screenshot 2019-09-16 at 11 56 22](https://user-images.githubusercontent.com/12841157/64945554-7362ad80-d879-11e9-8d43-dfc1dc6fc00d.png)

[...]

PR #951, opened 2019-08-07
adds className to MaterialTable props

pak11273:
## Related Issue
#766

## Description
Adds the prop className to MaterialTable so people can use className instead of style prop

* MaterialTable Component

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -456,7 +456,7 @@ export default class MaterialTable extends React.Component {
 
     return (
       <DragDropContext onDragEnd={this.onDragEnd}>
-        <props.components.Container style={{ position: 'relative', ...props.style }}>
+        <props.components.Container className={props.className} style={{ position: 'relative', ...props.style }}>
           {props.options.toolbar &&
             <props.components.Toolbar
               actions={props.actions}


src/material-table.js
```diff
@@ -456,7 +456,7 @@ export default class MaterialTable extends React.Component {
 
     return (
       <DragDropContext onDragEnd={this.onDragEnd}>
-        <props.components.Container style={{ position: 'relative', ...props.style }}>
+        <props.components.Container className={props.className} style={{ position: 'relative', ...props.style }}>
```
skrhlm: I thought of doing this as well, adding the  className prop to make the whole thing more similar to Mui. But in this specific case you _could_ just supply a custom Container component instead. :)

pak11273: I didn't think about overriding the Container component.  Maybe this should be stated in the styling section in the documentation.

pak11273: Just another thought.  If someone were using several material tables in their project, it would make more sense to have a class for styling.

skrhlm: Ah, I agrew with you. It would make a lot of sense to swap most of the custom component props for mui overrides instead! Also, I usually just put a thin wrapper component around MaterialTable, and expose that to the rest of the app with all my preferred settings :)

PR #959, opened 2019-08-08
Fix calculated width of Actions column to exclude hidden actions

dspatoulas:
## Description
The calculated width for the actions column includes hidden actions, which results in the column being too wide and a noticeable gap at the end of the column in the rows where the hidden actions would have been displayed.  This filters out the hidden columns from the calculation.


parventures:
I'm still facing this problem in the latest version. Having hidden columns introduces gaps in the actions column

Issue #964, opened 2019-08-08
initialPage property broken for remote data


bengennaria:
Using a remote data query and setting the `initialPage` to anything other than 0 in the options results in the "No Records to Display" message. `query.page === 0` when the the query is processed regardless of the value in `initialPage`. The api endpoint returns the same data as when initialPage is set to 0. If `initialPage === 0` the table correctly renders the query result.

**To Reproduce**
Set `initialPage` > 0 and attempt to load remote data. 

**Expected behavior**
`query.page` should be set to the `initialPage` value before the query is executed and the resulting data should be displayed.

```
                    options={{
                        pageSize: 25,
                        pageSizeOptions: [5, 10, 25, 50, 100, 200, 500],
                        filtering: true,
                        debounceInterval: 500,
                        initialPage: 5,
                    }}
                    data={query => new Promise((resolve, reject) => {
                            //console.log(JSON.parse(JSON.stringify(query)));
                            let url = '/api/entries'
                            let query_string = '?per_page=' + query.pageSize
                            query_string += '&page=' + (query.page+1)
                            query_string += '&orderby=' + query.orderBy.field
                            query_string += '&orderdir=' + query.orderDirection
                            query.filters.forEach(filter => {
                                query_string += '&filter[' + filter.column.field + ']=' + filter.value
                            });
                            if(query.search.length > 0)
                            {
                                query_string += '&search=' + query.search
                            }
                            url += query_string

                            fetch(url)
                                .then(response => response.json())
                                .then(result => {
                                    resolve({
                                        data: result.data,
                                        page: result.page - 1,
                                        totalCount: result.total,
                                    })
                                })
                        })
```

patrickp-dev:
Have opened a [PR](https://github.com/mbrn/material-table/pull/2007) to address the issue. Awaiting reviews/comments

allysonjp715:
Commenting so stale does not close the bug

PovilasKud:
I have the same issue and waiting for that PR to be merged :)

leandroz:
Same here.

andreyFernandoSoares:
I have the same problem here.

andreyFernandoSoares:
Here's one way.

`const curPage = localStorage.getItem('PAGE_EMPLOYEE_COMPANY');
  var intervalPage = null;
  let auxCount = 1;

  useEffect(() => {
    intervalPage = setInterval(() => {
      if (curPage != 0)
[...]

PR #974, opened 2019-08-10
support setting the initialPage option in conjunction with remote data 

anasofiapaixao:
Fixes issue #964 

This PR adds support to setting `intialPage` in conjuction with a remote data source.
It also adds a `postinstall` hook to allow getting the dependency directly from the git repository via npm or yarn. I added some additional entries in the in-memory data demo so that the demo showcases working pagination for both in-memory and remote data examples.

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -34,7 +34,7 @@ export default class MaterialTable extends React.Component {
           })),
         orderBy: renderState.columns.find(a => a.tableData.id === renderState.orderBy),
         orderDirection: renderState.orderDirection,
-        page: 0,
+        page: calculatedProps.options.initialPage || 0,
         pageSize: calculatedProps.options.pageSize,
         search: renderState.searchText,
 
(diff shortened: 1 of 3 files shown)


src/material-table.js
```diff
@@ -440,7 +440,7 @@ export default class MaterialTable extends React.Component {
                 SelectProps={{
                   renderValue: value => <div style={{ padding: '0px 5px' }}>{value + ' ' + localization.labelRowsSelect + ' '}</div>
                 }}
-                page={this.isRemoteData() ? this.state.query.page : this.state.currentPage}
+                page={this.isRemoteData() ? (Math.min(this.state.query.page, this.state.query.totalCount)) : this.state.currentPage}
```
mbrn: This could be a bug. query.totalCount is not count of pages. It is count of total rows.

src/material-table.js
```diff
@@ -536,7 +536,7 @@ export default class MaterialTable extends React.Component {
                         actions={props.actions}
                         components={props.components}
                         icons={props.icons}
-                        renderData={this.state.renderData}
+                        renderData={this.isRemoteData() ? this.state.data : this.state.renderData}
```
mbrn: Why do you set data directly to renderData?

mbrn:
I am closing this pr for now. You can reopen it after rebasing and change that i requested.

PR #996, opened 2019-08-16
 create new props  "onChangeColumnGrouped" 

eaparango:
Create new props "onChangeColumnGrouped" and add to types def typescript, Fix on Drop Column group to headers send event "onGroupRemoved"

### DIFF
--- a/src/prop-types.js
+++ b/src/prop-types.js
@@ -165,6 +165,7 @@ export const propTypes = {
   onChangeRowsPerPage: PropTypes.func,
   onChangePage: PropTypes.func,
   onChangeColumnHidden: PropTypes.func,
+  onChangeColumnGrouped: PropTypes.func,
   onOrderChange: PropTypes.func,
   onRowClick: PropTypes.func,
   onTreeExpandChange: PropTypes.func,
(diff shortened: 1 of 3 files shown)


src/prop-types.js
```diff
@@ -165,6 +165,7 @@ export const propTypes = {
   onChangeRowsPerPage: PropTypes.func,
   onChangePage: PropTypes.func,
   onChangeColumnHidden: PropTypes.func,
+  onChangeColumnGrouped: PropTypes.func,
```
mbrn: Please rename this event as onGroupOrderChanged

types/index.d.ts
```diff
@@ -27,6 +27,7 @@ export interface MaterialTableProps<RowData extends object> {
   onColumnDragged?: (sourceIndex: number, destinationIndex: number) => void;
   onOrderChange?: (orderBy: number, orderDirection: ("asc" | "desc")) => void;
   onGroupRemoved?: (column:Column<RowData>, index:boolean) => void;
+  onChangeColumnGrouped?: (column:Column<RowData>, index:boolean) => void;
```
mbrn: index should be number

mbrn:
I am closing this pr for now. You can reopen it after rebasing and change that i requested.

Issue #1002, opened 2019-08-19
Missing Aria Label Types for Pagination


dominikkaegi:
**Describe the bug**
TS Type definition of Localization Pagination is missing types for the aria-labels e.g. `firstAriaLabel`

**To Reproduce**
1. Import Localization Type
2. Use it on an object and try to use firstAriaLabel

```js
import { Localization} from 'material-table';

const localization: Localization = {
  pagination: {
    firstAriaLabel: 'Erste Seite',
  },
};
```
```
Error
Object literal may only specify known properties, and 'firstAriaLabel' does not exist in type
```

**Expected behavior**
The aria labels should be defined as types and throw no tsc errors.

**Additional context**
The existing aria-label types can be found in the [docs](https://material-table.com/#/docs/features/localization).

jayarjo:
PRs are welcome.

dominikkaegi:
We can close it. I already created a PR and it is already merged 👍

PR #1003, opened 2019-08-19
Add aria-label type definition to localization interface

dominikkaegi:
## Related Issue
Related Issue  [`#1002`](https://github.com/mbrn/material-table/issues/1002)

## Description
The goal is to add the missing types for the aria-labels of the pagination object within the localization interface.

## Related PRs
None

* The Type definitions index.d.ts

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -247,11 +247,15 @@ export interface Localization {
   };
   pagination?: {
     firstTooltip?: string;
+    firstAriaLabel?: string;
     previousTooltip?: string;
+    previousAriaLabel?: string,
     nextTooltip?: string;
+    nextAriaLabel?: string,
     labelDisplayedRows?: string;
     labelRowsPerPage?: string;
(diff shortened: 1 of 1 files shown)


jayarjo:
@dominikkaegi  I'm not that strong in Aria related stuff, but why would you need to have them as options, do they differe depending on the context, can't they be hardcoded?

dominikkaegi:
@jayarjo The options to pass on the aria labels already exists and are also documented [here](https://material-table.com/#/docs/features/localization). What I added was the type definitions for the aria labels for TS. Therefore I'm not really changing anything about the implementation, I'm just adding the missing types.

I'm not sure if I understand what you mean by hardcoding them. Overall I don't think you want to hardcode them because if you have multiple languages you support, you want to dynamically pass on the translated aria-label for a given language.

jayarjo:
@dominikkaegi thanks for clarification.

Issue #976, opened 2019-08-12
Specify totalCount without remote data


m5tt:
We are handling our data fetching & paginating manually (through Redux) and passing it to MaterialTable without a promise:

```
<MaterialTable
  ...
  title='Members'
  data={props.members}
/>
```

However, it appears that MaterialTable gets the total count from this array and doesn't let us specify. 

[This issue](https://github.com/mbrn/material-table/issues/281) had the same problem, and `remoteData` was added to allow the user to specify the `totalCount`, through a Promise.

However we have _zero_ need for using a Promise, and would like to pass `totalCount` straight to MaterialTable. Additionally a Promise doesn't work for us either, because the data is empty on the first render, and MaterialTable does not re call the Promise on data change. 

For example, this could be implemented in multiple ways:

Passing it as a normal prop:

```
<MaterialTable
  ...
  title='Members'
  data={props.members}
  totalCount={props.memberCount}
/>
```

or formatting data as an object, ie the object that Promise is supposed to return:

```
<MaterialTable
  ...
  title='Members'
  data={ { data: props.members, totalCount: props.memberCount, page: 0}
/>
```

ruthwe:
i have the exec same question, if it is possible to give data as array as specified, (
data | Array or func for remote data
) how do i give the total count?

tried also on pagination and didn't work 
     pagination:{
                labelRowsPerPage: this.state.rowsPerPage,
                page:page,
                totalCount:totalCount
            }

jtesch:
Using remote data will also not work with Apollo as it will just go into infinite loop. Prefer to pass totalCount through options.

ibandominguez:
Hi there, 

I was dealing with the same issue and I finally managed to overwrite de Pagination component and handle the events on my Container so I can dispatch redux actions.

I works well so far.

I used this component: https://material-ui.com/api/table-pagination/
and found about component overwriting here: https://material-table.com/#/docs/features/component-overriding

Hope it helps someone.

Cheers

FujiHaruka:
It seems to be resolved by https://github.com/mbrn/material-table/pull/795.

anmol5varma:
Can you help me with how to use totalCount with remote data?

azakharo:
> It seems to be resolved by #795.

Yes, it works, but not described in the documentation. 
Just specify simultaneously 2 props:
* page (current page)
* totalCount

```
            <MaterialTable
              ...
              data={data}
              page={page}
              totalCount={count}
              ... 
            />
[...]

PR #795, opened 2019-07-04
feat: foreign handling of pagination counters

spurreiter:
## Related Issue
None

## Description
We would like to manage remote data fetching outside the offered Promise based approach by making use of flux/redux like patterns.
For this we need to manage the pagination counters on our own independently of the given `data.length`. Usually {skip, limit, count} response parameters would give the page number and the totalCount.
This PR offers two new properties on `MaterialTable`:

- `prop.page` the number of the page the data-set is related
- `prop.totalCount` the total count of possible data sets

Only if both props are defined, the current behaviour of `MaterialTable` is changed such that the pagination values are managed outside the component. 

The change proposed is fully backwards compatible.

## Related PRs
None

## Impacted Areas in Application
List general components of the application that this PR will affect:

* MaterialTable.TableFooter

## Additional Notes

PR #1031, opened 2019-08-26
Add totalCount as a prop for non-remote data

kmiracle86:
## Related Issue
#976 

## Description
Allow passing a `totalCount` prop when not using `remoteData`. This will help Redux users that are using remote data but not with the `remoteData` option.

* MaterialTable component
* PropTypes

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -87,8 +87,14 @@ export default class MaterialTable extends React.Component {
     this.setState(this.dataManager.getRenderState());
   }
 
+  getTotalCount() {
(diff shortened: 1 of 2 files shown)


m5tt:
How do this compare to this - https://github.com/mbrn/material-table/pull/795

It's quite old now, but i dont know why it was never merged. It also exposes the `page` prop which is necessary if not using Promises.

jayarjo:
Closing as duplicate of #795. I think we will find time to merge all of these tiny fixes finally these days.

Issue #534, opened 2019-05-02
Editable Currency type render "ok"


icharge:
**Describe the bug**
I'm using Editable field data and set type as `currency` and then in inline editing **ok** is show up!.

**To Reproduce**
1. Set Columns and field type as `currency`.
2. Define editable `onRowUpdate` callback.
3. Run & click edit icon.
4. "ok" will show up.

**Expected behavior**
Inline Textbox for currency format.

**Screenshots**
![image](https://user-images.githubusercontent.com/1619007/57052929-c36b1a00-6cb4-11e9-8265-75b636c1ea82.png)

**Desktop (please complete the following information):**
 - OS: Any
 - Browser Any

bobctr:
As a workaround for now, you can overwrite the editComponent in the column definition:

```
{
      title: "Value",
      field: "value",
      type: "currency",
      editComponent: props => (
        <input
          type="numeric"
          value={props.value}
          onChange={e => props.onChange(e.target.value)}
        />
      )
    }
```

eddiemonge:
Its because there is no currency field definition at the moment: https://github.com/mbrn/material-table/blob/v1.50.0/src/components/m-table-edit-field.js#L126

mforman1:
> As a workaround for now, you can overwrite the editComponent in the column definition:
> 
> ```
> {
>       title: "Value",
>       field: "value",
>       type: "currency",
>       editComponent: props => (
>         <input
>           type="numeric"
>           value={props.value}
>           onChange={e => props.onChange(e.target.value)}
>         />
>       )
> }
> ```

@bobctr Thanks for the workaround, do you have any idea how can I make it similar to the other fields?

![image](https://user-images.githubusercontent.com/39950157/66707891-b74db300-ed50-11e9-8e30-24e909a66fd4.png)

Domino987:
Cannot be reproduced anymore. IF this is still relevant, feel free to reopen this issue.

PR #1056, opened 2019-08-31
Fix Currency field Issues - Edit and alignment

yonihorwitz:
## Related Issue
[#534](https://github.com/mbrn/material-table/issues/534) - Editable Currency type render "ok"

## Description
When editing a currency field, instead of returning "ok", show a right-aligned textField.  Also, right align currency field render when not in Edit mode.

* Affects MTableEditField and MTableCell

## Additional Notes
I attempted a previous PR for the same issue earlier - this is a new and improved version.  Still though, I'm new to working with GitHub so would appreciate any feedback.

### DIFF
--- a/src/components/m-table-edit-field.js
+++ b/src/components/m-table-edit-field.js
@@ -123,7 +123,20 @@ class MTableEditField extends React.Component {
   }
 
   renderCurrencyField() {
-    return "ok";
+    return (
+      <TextField
+        {...this.getProps()}
+        placeholder={this.props.columnDef.title}
+        value={this.props.value === undefined ? '' : this.props.value}
+        onChange={event => this.props.onChange(event.target.value)}
+        inputProps={{
+          style: {
+            fontSize: 13,
+            textAlign: "right"
+          }
+        }}
+      />
+    );
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


src/components/m-table-edit-field.js
```diff
@@ -123,7 +123,20 @@ class MTableEditField extends React.Component {
...
+    return (
+      <TextField
+        {...this.getProps()}
+        placeholder={this.props.columnDef.title}
+        value={this.props.value === undefined ? '' : this.props.value}
+        onChange={event => this.props.onChange(event.target.value)}
+        inputProps={{
+          style: {
+            fontSize: 13,
+            textAlign: "right"
```
eddiemonge: should copy the `style={{ float: 'right' }}` like from TextField above

src/components/m-table-edit-field.js
```diff
@@ -123,7 +123,20 @@ class MTableEditField extends React.Component {
   }
 
   renderCurrencyField() {
-    return "ok";
+    return (
+      <TextField
+        {...this.getProps()}
+        placeholder={this.props.columnDef.title}
+        value={this.props.value === undefined ? '' : this.props.value}
```
eddiemonge: missing `type='number'`

mforman1:
Hi,
Any update on the above? Can that be merged?
Thanks.

mforman1:
Thanks! Any idea when the next version will be released?

Issue #760, opened 2019-06-25
Tree data does not indent


jakeleventhal:
**Describe the bug**
Tree data does not indent when dropped down

**To Reproduce**
1. Create tree data with multiple dropdowns
2. Expand each row

**Expected behavior**
The title for each dropdown should be indented to show that it is a child of the parent - just like how the arrows indent. The behavior is also correct on grouped tables

**Screenshots**
<img width="376" alt="Screen Shot 2019-06-25 at 7 16 00 PM" src="https://user-images.githubusercontent.com/15133828/60139915-c3037280-977d-11e9-9a0b-a66bff464503.png">

jakeleventhal:
It should be a prop because sometimes this behavior is desired

waisbek:
Any update about this issue, please?

benjaminvdb:
I'm also looking for a solution!

PS: material-table is awesome 😃

emrekara37:
Hi everyone 
We will fix asap

dnn5b:
There's a [pull request](https://github.com/mbrn/material-table/pull/1057) for this. Any feedback?

blshort43:
Any solution to this yet?

jakeleventhal:
not stale!

Phillip-Kemper:
not stale. I would greatly appreciate if the Pull Request could be merged in as I would love to continue using material-table and do prefer it over lower level table libraries..

asharafshahi:
This could be accomplished with a simple custom column render function.  Just render a <div> and pass in a left padding according to the depth of the tree the row is at

jakeleventhal:
no stale

PhilObi:
Has there been any solutions for this?

lpandzic:
```
 columns={[
                                {
                                    title: 'Service Unique Name',
                                    type: 'string',
                                    field: 'serviceUniqueName',
                                    editable: 'never',
                                    cellStyle: (rowDataArray, rowData) => {
                                        const paddingLeft = 30 * rowData.level;
                                        console.log(paddingLeft)
                                        return ({ paddingLeft: paddingLeft} as React.CSSProperties)
                                    }
                                },
```

lpandzic:
Here's my solution:

```
 columns={[
    {
        title: 'FooBar',
        type: 'string',
        field: 'fooBar',
        cellStyle: (rowDataArray, rowData) => {
            const paddingLeft = 30 * rowData.level;
            console.log(paddingLeft)
[...]

PR #1057, opened 2019-08-31
Table indents

yonihorwitz:
## Related Issue
[Tree data does not indent](https://github.com/mbrn/material-table/issues/760)

## Description
Currently indentation only shows for the chevron icons and for selection checkboxes.  This PR adds two options to MaterialTable:  indentActionsCell and indentFirstDataCell.  Setting either of these to true will indent the corresponding components.

* MTableBodyRow

## Additional Notes
Only my second PR - any feedback welcome.

### DIFF
--- a/src/components/m-table-body-row.js
+++ b/src/components/m-table-body-row.js
@@ -17,12 +17,20 @@ export default class MTableBodyRow extends React.Component {
       .sort((a, b) => a.tableData.columnOrder - b.tableData.columnOrder)
       .map((columnDef, index) => {
         const value = this.props.getFieldValue(this.props.data, columnDef);
+        const indentStyle = (this.props.options.indentFirstDataCell && index == 0)?(
+          size === 'medium' ? {
+            paddingLeft: this.props.level * 9
+          } : {
+            paddingLeft: 5 + this.props.level * 9
+          }
+        ):{};
         return (
           <this.props.components.Cell
             size={size}
             icons={this.props.icons}
             columnDef={columnDef}
             value={value}
+            style={indentStyle}
             key={"cell-" + this.props.data.tableData.id + "-" + columnDef.tableData.id}
             rowData={this.props.data}
           />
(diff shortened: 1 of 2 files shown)


benjaminvdb:
> The solution looks good. But I personally would prefer a bigger indention (e.g. 15px).

Perhaps make the indent size can be passed as a prop?

yonihorwitz:
Agreed - I'll make the changes once I've delivered on a timeline tomorrow...

General GitHub question - why am I making these PRs if it looks like the owner is AWOL?

src/components/m-table-body-row.js
```diff
@@ -17,12 +17,20 @@ export default class MTableBodyRow extends React.Component {
       .sort((a, b) => a.tableData.columnOrder - b.tableData.columnOrder)
       .map((columnDef, index) => {
         const value = this.props.getFieldValue(this.props.data, columnDef);
+        const indentStyle = (this.props.options.indentFirstDataCell && index == 0)?(
+          size === 'medium' ? {
+            paddingLeft: this.props.level * 9
```
jakeleventhal: I dont think this is the best way to do it.

Consider a nesting 3-4 layers deep. you will want the indentation to still work

jayarjo:
What would you guys prefer having it like suggested here, or simply - some classNames exposed for overriding?

benjaminvdb:
> What would you guys prefer having it like suggested here, or simply - some classNames exposed for overriding (in generic material-ui-like way)?

If material-ui has a common way of doing things, then I'd go for that, unless the indentation width would be located in a deeply nested style property, in which case a component prop would be more convenient.

I reckon quite a few people would be interested in this option, so we could include the indentation width in an example in the documentation.

yonihorwitz:
Changed the options to accept numbers (as opposed to boolean), which are then applied to the indents as pixels.  Apologies for the delay.

OnkelTem:
> We don't want to add more style props to MaterialTable. We will add more effective styling options.

That of course is good decision, but how to indent then **now**?

yonihorwitz:
I haven't been able to figure out how to get it exactly like I want, but this is what I have for now.

`rowStyle: (rowData, index, level) => ({
                  backgroundColor:
                    rowData.review_required === 1
                      ? "#e2affa"
                      : level > 0
                      ? "#EEEEEE"
                      : "white",
                  "&:nth-child(3)": {
                    marginLeft: level * 9
                  }
                })`

The rowStyle property takes a _level_ parameter.  You can use it to highlight child rows.  It doesn't work as well when you have multiple levels.

I tried using nth-child as shown above but **it doesn't work.**  I don't know why, but the parser doesn't seem to like it, even though all my Googling says it should work...  If you can figure it out, let me know.

OnkelTem:
Let me share my solution. It generate indents for 10 levels:

```js
const useStyles = makeStyles(theme => {
  const map = Array.from(Array(10).keys()).reduce((map, i) => {
      map[`& .MuiTable-root .MuiTableRow-root[level="${i + 1}"]>td:nth-child(2)`] = {
        paddingLeft: theme.spacing(2 + 4 * (i + 1)),
      };
      return map;
    }, {});
  return {
    table: {
      ...map
    },
    // other styles
  }
});

...

export default function YourComponent() {
  const classes = useStyles();
  return (
      <div className={classes.table}>
        <MaterialTable
          ...
```

Issue #739, opened 2019-06-20
Error : getFieldValue is not a function when editing a child (Tree Data)


jessy-bgl:
**Describe the bug**
When I click on the "edit" action button of a child row of a Tree Data table **with parameter _editable: "onAdd"_ on one of the columns**, I get the error : 

> Uncaught TypeError: _this2.props.getFieldValue is not a function

**To Reproduce**
1. Create a Tree Data table **with the parameter editable: "onAdd" on (at least) one column**
2. Add a row
3. Add one or more children to this row
4. Click on the Edit action button
5. See the error

**Expected behavior**
Ability to edit the fields of the row.

**Screenshots**
![image](https://user-images.githubusercontent.com/51905435/59837458-93073a00-934d-11e9-876e-f2d8f266f147.png)

> Clicking on the edit action button of the row "groupe10" :

![image](https://user-images.githubusercontent.com/51905435/59837473-9995b180-934d-11e9-97c7-bba12122d304.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser : Chrome
 - Version 74.0.3729.169

Xzas22:
Having the same error message pop up. Everything appears to work correctly, but the React error is concerning.

theodore-koch:
Same here when setting editable param to 'never'

foramshah1112:
Getting the same error on clicking Add Icon for the first time after page loads, when using editComponent, later the error goes away on it's own.

jayarjo:
Same here. Also would be great to have column drag&drop logic configurable, so that it could be switched off if required (observing the same `react-beautiful-dnd` error here).

HazilMohamed:
Is there any fix available right now??

cristiadis:
Same issue here. Adding "editable: 'never'" to a column causes the issue when clicking on Edit.

RamilMamedo:
any updated on this? ☹️

Tobbe129:
Please open this again, its not fixed!

PR #1058, opened 2019-08-31
getFieldValue Fix: Pass props to EditRow

yonihorwitz:
## Related Issue
#739 

## Description
1)  Props are passed down to components.Row on 308, but aren't similarly passed to components.EditRow on 290.  This PR changes that so that props are passed to EditRow as well, so that getFieldValue is passed to the edit field.  

2)  Step #1 results in a couple of browser warnings.  In MTableEditRow, filter out the two properties from the props object that cause them.

* MTableEditRow
* MTableBodyRow

## Additional Notes

### DIFF
--- a/src/components/m-table-body-row.js
+++ b/src/components/m-table-body-row.js
@@ -288,6 +288,7 @@ export default class MTableBodyRow extends React.Component {
             if (data.tableData.editing) {
               return (
                 <this.props.components.EditRow
+                  {...this.props}
                   columns={this.props.columns.filter(columnDef => { return !columnDef.hidden })}
                   components={this.props.components}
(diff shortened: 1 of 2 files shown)


src/components/m-table-body-row.js
```diff
@@ -288,6 +288,7 @@ export default class MTableBodyRow extends React.Component {
             if (data.tableData.editing) {
               return (
                 <this.props.components.EditRow
+                  {...this.props}
```
mbrn: Passing all props is not required. Please pass just getFieldValue.

yonihorwitz:
Looks like the change has since been made by another PR.  Closing.

PR #1059, opened 2019-09-02
fix double click handler in renderDefaultActions

Paktusin:
## Related Issue
when click on formCotrolLabel clickHandler call twice


mikaelbr:
Hi. Sorry, I didn't see this PR initially. I added a different PR (#1151) also solving this, but in a different way. I kind of think the markup changes here reduces the semantic meaning of the HTML, and is a more complicated solution than is necessary.

I added a more detailed explanation in the #1151 under `Additional Notes`.

mbrn:
#1151 will be merge for this problem.

Issue #765, opened 2019-06-26
Fixed Columns


francesco-clementi:
Is there a way to fix the position of the first column in a table with over 40 columns?

dnn5b:
Created a [PR](https://github.com/mbrn/material-table/pull/1076) for this feature. What do you think about it?

francesco-clementi-bip:
Hi dnn5b,
actually I was looking for another kind of fixed columns. 

As shown in the attachment, I'd like to fixed the column when I use the stick on the down.

I'd like that when I'm scrolling right I can be able to still see the first columns.

![image](https://user-images.githubusercontent.com/50021302/64433112-21be6400-d0be-11e9-92a6-79c32f252119.png)

ShevaShen:
I was able to fix the first column (action column) by applying some CSS.

`const useStyles = makeStyles({
  stickyActionsColumn: {
    '& table:first-child': {
      '& tr': {
        '& td:first-child, th:first-child': {
          backgroundColor: '#f5f5f5',
          position: 'sticky',
          left: 0,
          zIndex: 999
        },
        '& th:first-child': {
          zIndex: 9999
        }
      }
    }
  }
});`

francesco-clementi-bip:
@ShevaShen thanks sheva, your help is really appreciated,.
I managed to fixed the header of the actions column.
Could it be possibile to fix not only the header, but the whole column?

ShevaShen:
@francesco-clementi-bip if you apply the style to the parent div of the table, this should fix the whole first column

codenamethanos:
https://material-table.com/#/docs/features/fixed-columns 
Seems to be solved?

ShevaShen:
right, saw it in the new release https://github.com/mbrn/material-table/issues/43

javed-ahmad:
looking for same solution with sticky header any suggestion?

PR #1076, opened 2019-09-06
Feature/fix columns

dnn5b:
## Related Issue
https://github.com/mbrn/material-table/issues/765

## Description
Added a `fixed` property to column definition. If this property is true a column is not moveable to another position. Other columns are not affected by this property. 
It's useful if e.g. the first column of a table has to be fixed at the position.

branch | PR
------ | ------
Branch| [https://github.com/dnn5b/material-table/tree/feature/fix-columns]()

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -47,7 +47,7 @@ class App extends Component {
       { id: 6, name: 'A6', surname: 'C', isMarried: true, birthDate: new Date(1989, 1, 1), birthCity: 34, sex: 'Female', type: 'child', insertDateTime: new Date(2018, 1, 1, 12, 23, 44), time: new Date(1900, 1, 1, 14, 23, 35), parentId: 5 },
     ],
     columns: [
-      { title: 'Adı', field: 'name', filterPlaceholder: 'Adı filter' },
+      { title: 'Adı', field: 'name', filterPlaceholder: 'Adı filter', fixed: true },
       { title: 'Soyadı', field: 'surname', initialEditValue: 'test' },
       { title: 'Evli', field: 'isMarried', type: 'boolean' },
       { title: 'Cinsiyet', field: 'sex', disableClick: true, editable: 'onAdd' },
--- a/src/components/m-table-header.js
+++ b/src/components/m-table-header.js
@@ -18,7 +18,7 @@ export class MTableHeader extends React.Component {
         let content = columnDef.title;
 
         if(this.props.draggable) {
-          content = (
+          content = (columnDef.fixed ? <>{columnDef.title}</> : 
             <Draggable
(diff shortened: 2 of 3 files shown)


src/components/m-table-header.js
```diff
@@ -18,7 +18,7 @@ export class MTableHeader extends React.Component {
         let content = columnDef.title;
 
         if(this.props.draggable) {
-          content = (
+          content = (columnDef.fixed ? <>{columnDef.title}</> :
```
xiaomaini: Thoughts on changing this from `fixed` to `draggable`?

src/components/m-table-header.js
```diff
@@ -18,7 +18,7 @@ export class MTableHeader extends React.Component {
         let content = columnDef.title;
 
         if(this.props.draggable) {
```
xiaomaini: Maybe you can lift the logic in line 21 up if you rename to `draggable`, example: 

```
if(this.props.draggable && columnDef.draggable) {
  ...
}
```

src/prop-types.js
```diff
@@ -38,6 +38,7 @@ export const propTypes = {
     grouping: PropTypes.bool,
     headerStyle: PropTypes.object,
     hidden: PropTypes.bool,
+    fixed: PropTypes.bool,
```
xiaomaini: rename to `draggable` and add a default prop to `true`

demo/demo.js
```diff
@@ -47,7 +47,7 @@ class App extends Component {
       { id: 6, name: 'A6', surname: 'C', isMarried: true, birthDate: new Date(1989, 1, 1), birthCity: 34, sex: 'Female', type: 'child', insertDateTime: new Date(2018, 1, 1, 12, 23, 44), time: new Date(1900, 1, 1, 14, 23, 35), parentId: 5 },
     ],
     columns: [
-      { title: 'Adı', field: 'name', filterPlaceholder: 'Adı filter' },
+      { title: 'Adı', field: 'name', filterPlaceholder: 'Adı filter', fixed: true },
```
xiaomaini: rename to `draggable`

CoolCyberBrain:
Should also add typescript definition `fixed?: boolean;` to `export interface Column<RowData extends object>` in types/index.d.ts right after `filterCellStyle?: React.CSSProperties;`

jayarjo:
I agree, idea is decent, however adding new verb - not good. Vote for `draggable`. Mind to change @dnn5b?

mbrn:
Wait too much.

marianungureanu:
Is there any solution for this. I want to use draggable feature, but I want that some columns to be fixed on their position.

Issue #573, opened 2019-05-12
Add and control table footer


flyingangel:
Not talking about the footer with pagination. But the last row where you can display the sum or the percentage of one column.

The row must be fixed when navigating with pagination, must not have any empty space between this footer row and the last row of a table

We can set footer for all or just one chosen column but the number of declared footer must be the same as header columns

The footer, by default, use a **sum** function if the header of the table is of type **number**, or **percentage** with an option to specify precision after **,** or **.** (local dependent). A custom rendering function is needed, receiving the entire table data as function parameter.

If any row in a table is update or through onRowUpdate/Delete/Add, footer will be re-rendered

https://datatables.net/examples/advanced_init/footer_callback

mbrn:
This is my roadmap. I will add a summarize row feature asap.

iherger:
I am also looking for such a sum row at the end. That would really be cool, and would make material-table even better than it already is!

codenamethanos:
I also want this feature really badly.  
Is it atleast possible to make a custom row? And then render some data into it and use it as a workaround with current features?

radmonteiro:
any news when will this be available?

mindaugasnakrosis:
@mbrn 

> This is my roadmap. I will add a summarize row feature asap.

What is the status?

PR #1078, opened 2019-09-06
Footer implemented - activated by 'footerData' prop

seanharr11:
## Related Issue
#573 

## Description
This PR adds a `footerData` prop to `MaterialTable`, and implemented the `MtTableFooter` component. The `footerData` prop is of type `{[key in keyof RowData]: string | number | React.ReactElement | null`. The **keys** in `footerData` will be used to align the **values** with the correct columns in the footer cells, taking into account "placeholder cells" for features such as grouping, actions, detailPanel, etc.  If `footerData === null` the footer will not be rendered.

## Related PRs
Nada

* `MaterialTable`
    * Adds `footerData` prop to component API.
 
_Most of this is new functionality, not affecting old code_

## Additional Notes
I think this is a *good enough* step towards a long-term `Footer` implementation.  Most users will be able to access this API quite easily, by passing an object via `footerData` that has the same shape as the rest of the data passed to the table. Future improvements could include:
1. Automatic aggregations (i.e. avg., sum., stdev., etc.)
2. Multiple footer rows
3. A 'label' for a footer row (_often times you want to do something like *Avg:* `123` | `10.6` | `9.5`|_)

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -8,6 +8,7 @@ import MTableCell from './m-table-cell';
 import MTableEditRow from './m-table-edit-row';
 import MTableEditField from './m-table-edit-field';
 import MTableFilterRow from './m-table-filter-row';
+import MTableFooter from './m-table-footer';
 import MTableHeader from './m-table-header';
 import MTablePagination from './m-table-pagination';
 import MTableSteppedPagination from './m-table-stepped-pagination';
@@ -24,6 +25,7 @@ export {
   MTableEditRow,
   MTableEditField,
   MTableFilterRow,
+  MTableFooter,
   MTableHeader,
   MTablePagination,
   MTableSteppedPagination,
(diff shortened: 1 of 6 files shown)


jayarjo:
More components to override @mbrn :)

Amzd:
Is there a way to use this in my project before it is merged in?

Amzd:
Notes after forking this and trying it out for myself:

- onRowClick is applied to the footer too, this should either have onFooterClick or just not be handled at all.
- When using detailPanel the footer doesn't take into account the extra cell it has to move over (maybe for actions too, didn't test that)
- footerData takes in an object to show, this makes it hard to show totals when filters are applied (I couldn't figure out how to do this). I think the footerData type should be `(rows: RowData[]) => {...}` where the rowdata array is all the rows in the table (incl other pages, excl filtered out rows).

mbrn:
I think that we need a footer that has more features, such as sum, avg etc. But this footer implementation only show data sent. We can do better. Thanks @seanharr11

Issue #948, opened 2019-08-07
TypeScript types for custom actions are wrong


vinterdo:
I have custom action on table. Typescript requires function to have signature:
`public customAction(event: any, rowData: IRowData[])`
It is ok when selection is enabled, but when selection is disabled and action is executed on single row, rowData is just object of this row (not an array). So now i need to do @ts-ignore or my code dont work.

gianlucamateo:
The type for parentChildData is also wrong, as the function returns a single dataset and not an array.

adamTarr:
Also running into this issue. In a `action`, the `onClick` function is typed as 
`Action<RowData>.onClick: (event: any, data: RowData | RowData[]) => void`
This makes custom actions unusable in a TypeScript project unless the error is ignored with `//@ts-ignore`

PR #1081, opened 2019-09-07
Fix type definition for "parentChildData"

CoolCyberBrain:
## Related Issue
https://github.com/mbrn/material-table/issues/948#issuecomment-519097335

## Description
According to the [example for tree data](https://material-table.com/#/docs/features/tree-data). parentChildData should take a function that returns RowData not RowData[]

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -19,7 +19,7 @@ export interface MaterialTableProps<RowData extends object> {
   isLoading?: boolean;
   title?: string | React.ReactElement<any>;
   options?: Options;
-  parentChildData?: (row: RowData, rows: RowData[]) => RowData;
+  parentChildData?: (row: RowData, rows: RowData[]) => RowData | undefined;
   localization?: Localization;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -19,7 +19,7 @@ export interface MaterialTableProps<RowData extends object> {
   isLoading?: boolean;
   title?: string | React.ReactElement<any>;
   options?: Options;
-  parentChildData?: (row: RowData, rows: RowData[]) => RowData[];
+  parentChildData?: (row: RowData, rows: RowData[]) => RowData;
```
a-fortunato: Shouldn't it be `=> RowData | undefined` to be able to use `rows.find(a => a.id === row.parentId)` as in the [example](https://material-table.com/#/docs/features/tree-data)?
So find returns RowData if it finds a parent for the row and undefined if not.

CoolCyberBrain: Yep you're right. Just commited the change.

PR #979, opened 2019-08-13
Housekeeping - React upgrade + prettier + jest

skrhlm:
## Description
This PR aims to aid the workflow when contributing to this project. Bringing conformity to linting and testing.
- [Upgrade to the latest React version.](https://reactjs.org/blog/2019/08/08/react-v16.9.0.html)
- [Jest](https://jestjs.io) for testing and added two ( very simple ) tests for now.
- [Prettier](https://github.com/prettier/prettier) to enforce the linting style across the whole project.
- Broke out the data used by the demo application to allow this to be used in for example tests.
- I chose single quotes over double quotes since it seems as it's been used historically in this repo.
- No inner workings of the MaterialTable itself has been made, merely linted. 

## Impacted Areas in Application
* Push / commit workflow
* All files has been formatted using prettier, so basically everything.
* The demo application has been broken down into smaller files.
* Dotfiles ; `.eslintrc`, `.prettierrc.js`, `.babelrc`

## Additional Notes

If this is to work we need to add tests to more of the components : 
- [ ] m-table-action.js
- [ ] m-table-actions.js
- [ ] m-table-body-row.js
- [ ] m-table-body.js
- [x] m-table-cell.js
- [ ] m-table-edit-field.js
- [ ] m-table-edit-row.js
- [ ] m-table-filter-row.js
- [ ] m-table-group-row.js
- [ ] m-table-groupbar.js
- [ ] m-table-header.js
- [ ] m-table-pagination.js
- [ ] m-table-stepped-pagination.js
- [ ] m-table-toolbar.js
- [ ] m-table
- [x] utils

PR #1120, opened 2019-09-21
Revert "Housekeeping - React upgrade + prettier + jest"

mbrn:
Reverts mbrn/material-table#979


Domino987:
Why did this get reverted? What were the problems?

jayarjo:
@Domino987 it touches every line in the code basically, so there would have been conflicts in every PR. This should be merged. But last.

Issue #725, opened 2019-06-15
Initial Search Prop


iPwnPancakes:
Is it possible to specify a default search text as a prop? So far I have tried overriding the MTableToolbar component with a searchText prop as well as an onSearchChanged function to update the state, though it doesnt filter the data until after I have typed

a544jh:
I'm also looking for exactly this. I would be ready to implement this and open a MR.

mumispb:
I've been also trying to change the search placeholder text. As far as I understand, you could add a components prop to <MaterialTable /> to override the default Toolbar, like so: 

`components={{Toolbar: props => <MTableToolbar  {...props} searchPlaceholder="Your text"/>}}`

Or your could try to change localization prop, but I couldn't get any to work.

Overriding components: https://material-table.com/#/docs/features/component-overriding
Props: https://material-table.com/#/docs/all-props

mumispb:
Actually just found out that when you try to change localization prop, like this:

```
<MaterialTable
                localization={{
                    toolbar: {
                        searchPlaceholder: "Your search placeholder"
                    },
                }}
/ >
```

An error occurred, saying that body.deleteToolTip was undefined, so I gave it its default value, like so and it worked:

```
<MaterialTable
                localization={{
                    toolbar: {
                        searchPlaceholder: "Procurar"
                    },
                    body: {
                        deleteTooltip: "Delete"
                    }
                }}
/>
```

a544jh:
@mumispb This isn't was I was looking for. I'm looking to set search value prop that will actually apply the filter when the prop is changed. This is so that the search value could be loaded from a URL query param.

Setting `defaultFilter` in for a column in the `columns` prop works in the desired way (i.e. the filter gets applied when the prop is changed).

tobilarscheid:
Hi guys, any updates on this? I am looking for the same thing - a default search text I can pass as a prop that actually gets applied.

tobilarscheid:
For anyone interested - I have just opened a PR: #1137 1137

jayarjo:
@a544jh I'd recommend to open new issues for unrelated problems, do not mix topics please, it's confusing.

tobilarscheid:
I think this can be closed since #1225 was merged.

Issue #857, opened 2019-07-24
How to make a custom search in Mat-Table


sayokbose91:
I have a custom column which looks like this :

 { title: 'Email', field: 'email' },
            {
                title: 'skills',
                customFilterAndSearch: rd => this.CustomFilterMethod(rd),
                render: rowData => {
                    return rowData.skills.map((s, i) => (<Tooltip title={"Skill Level :" + rowData.skillScores[i]}><Chip avatar={<Avatar color="secondary">{rowData.skillScores[i]}</Avatar>} style={{ margin: 3 }} label={s} /></Tooltip>))
                },
            },

now how should i write a custom search on this custom column so that it only shows rows where the user has the specific skills?
I see there is a customFilterAndSearch option but i am nit sure what should i return for each rowdata....

api data model which i sue to bind my table looks like this: 

[
  {
    "name": "string",
    "email": "string",
    "id": "string",
    "skills": [
      "string"
    ],
    "skillScores": [
      int
    ],    
  }
]

schen3:
columns: [
    {
      title: `memberName (${36})`,
      field: 'memberName',
      filtering: true,
      **search: false**
    },
    { title: 'Work email', field: 'workEmail', filtering: false },
    { title: 'Personal email', field: 'personalEmail', filtering: false },

  ]

When you define your columns data, you can make specific column not searchable.

schen3:
You can close search for any column by setting column.searchable as false

mqasim1990:
@sayokbose91 [Possible Solution](https://github.com/mbrn/material-table/issues/1119#issuecomment-536701020)

raph90:
@mqasim1990 What query.search do you mean?

mqasim1990:
Hi @raph90,

```js
// Create a state
constructor() {
    super();
    this.state = {
      searchQuery: [],
      tableQuery: '',
    }
    this.tableRef = React.createRef();
  }

/**
   * Function to fetch all Data based on search query
   * Keep in mind **query** is the default parameter for tableData element. Which returns an array with required search      * values.
   */
  fetchData(clientId, query) {
    query.search = this.state.searchQuery;
    this.setState({ tableQuery: query });
    return authenticationService.getAllNotifications(clientId, query)
      .then((result) => {
        const tableData = {
          data: result.data.data.records,
          page: result.config.params.page - 1,
          totalCount: result.data.data.count,
        };
        return tableData;
      });
  }

[...]

PR #1137, opened 2019-09-29
Feature: Initial Search Text

tobilarscheid:
## Related Issue
#725 

## Description
This PR enables passing an initial search value as a prop to `<MaterialTable/>`. This can for example be used to deep link to a search result in the table.

* `data-manager.js` - has a new property carrying the initial search text
* `material-table.js` - needs to pass initial search text from props to data manger and trigger a search right after the component was mounted

## Additional Notes
I would have loved to update the respective documentation, but couldn't find material-table.com sources. If you can point me there, I will be happy to update.

### DIFF
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -158,6 +158,10 @@ export default class DataManager {
     this.currentPage = 0;
   }
 
+  setSearchInitialText(searchInitialText) {
+    this.searchInitialText = searchInitialText;
+  }
+
   changeRowEditing(rowData, mode) {
     if (rowData) {
       rowData.tableData.editing = mode;
@@ -474,6 +478,7 @@ export default class DataManager {
       pageSize: this.pageSize,
       renderData: this.pagedData,
       searchText: this.searchText,
+      searchInitialText: this.searchInitialText,
       selectedCount: this.selectedCount,
       treefiedDataLength: this.treefiedDataLength,
       treeDataMaxLevel: this.treeDataMaxLevel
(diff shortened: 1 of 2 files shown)


jayarjo:
I personally don't like that we keep adding properties to our config, which is already very big. @mbrn I think we should come up with a better way to pass to pass props to the child components, instead of passing them all the way down from the parent.

One of the approaches I suggested is simply have a way to override each and every child component via components prop.

Internally we could also make use of Context API.

Suggestions are welcome.

jayarjo:
Related: https://github.com/mbrn/material-table/issues/857

klappy:
@jayarjo I understand your concerns about not letting the passing of props be so broad and continue to grow. It is overwhelming. 

Is it possible to take this approach, or guide the PR to the current standard in the library? Then refactor the approach for a future major release?

If you do end up punting on this PR for a new approach, do you have a recommended workaround to accomplish the same functionality?

aleccaputo:
@tobilarscheid Before I go looking any further, i'm looking to "prefill" the search field from a value in state when a use navigates back to the table. I'm assuming that's what this pull accomplishes?

tobilarscheid:
@aleccaputo exactly that!

aleccaputo:
@tobilarscheid @jayarjo until this get's worked into master, is there any workaround to accomplish something similar. Placeholder doesn't work because it don't actually let you clear out the search filter that is pre-applied

src/utils/data-manager.js
```diff
@@ -158,6 +158,10 @@ export default class DataManager {
     this.currentPage = 0;
   }
 
+  setSearchInitialText(searchInitialText) {
+    this.searchInitialText = searchInitialText;
+  }
```
mbrn: Why you didn't use searchText.

tobilarscheid: Hi Mehmet, just modifying `searchText` didn't seem to be enough: I also need to notify `onSearchChange` - which then set's `searchText` on the `dataManager`. I did not want to interfere with this. If you say it's save to use, I can change it, but I did feel like it is cleaner when `searchText` is only modified from one place at a time.

mbrn:
Duplicate with #1225

PR #1145, opened 2019-10-01
Add approve to edit Component

Domino987:
## Related Issue
[Issuse](https://stackoverflow.com/questions/58099237/add-new-row-with-a-key-press-like-enter)

## Description
This PR adds the approve function to the column editComponent props so that the the developer can execute the approving of an edit manually or on a certain key hit. For example on enter.

The props of editComponent not contain the approveEdit function, which triggers `onEditingApproved`

* Edit Row


mbrn:
Hi @Domino987 ,

Thanks for contribution. But i will implement a validation like yup, formik. So i will close this pr.

Issue #1129, opened 2019-09-25
Allow date/datetime/time pickers to be localized.


ogenodisho:
**Is your feature request related to a problem? Please describe.**
Yes, I would like to localize the date/datetime/time pickers so that my users can view the data in their locale.

**Describe the solution you'd like**
Currently, the `m-table-edit-field` and `m-table-filter-row` javascript files contain `material-ui /pickers` that are wrapped with a `MuiPickersUtilsProvider` component. This component can be updated with the `locale` prop which can be supplied somehow by the user of this library so that the pickers become localized. Here is an example: https://material-ui-pickers.dev/localization/date-fns

**Describe alternatives you've considered**
I haven't thought of alternatives for the moment.

**Additional context**
None.

rommelmamedov:
Any updates on this issues? :)

sureshtidbit:
is this completed? @mbrn

PR #1149, opened 2019-10-01
Add locale to datepicker

Domino987:
## Related Issue
#1129 

## Description
This adds a localisation prop to pass a date-fns locale object within the body to enable datepicker localisation.
Like this: 
` import frLocale from 'date-fns/locale/fr';`
...

`localisation={{body: {locale: frLocale }}}`

This will result in a french datepicker.

image

* Body
* Edit Field
* Edit Row
* Filter Row
* index.d.ts

### DIFF
--- a/src/components/m-table-body.js
+++ b/src/components/m-table-body.js
@@ -47,7 +47,7 @@ class MTableBody extends React.Component {
             components={this.props.components}
             data={data}
             icons={this.props.icons}
-            localization={{ ...MTableBody.defaultProps.localization.editRow, ...this.props.localization.editRow }}
+            localization={{ ...MTableBody.defaultProps.localization.editRow, ...this.props.localization.editRow, dateTimePickerLocalization: this.props.localization.dateTimePickerLocalization }}
             key={index}
             mode={data.tableData.editing}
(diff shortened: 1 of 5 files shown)


types/index.d.ts
```diff
@@ -225,6 +225,7 @@ export interface Options {
 
 export interface Localization {
   body?: {
+    locale?: object; // The date-fns locale object applied to the datepickers
```
mbrn: Can you rename this as dateTimePickerLocalization, please?

Domino987: Done

src/components/m-table-body.js
```diff
@@ -47,7 +47,7 @@ class MTableBody extends React.Component {
             components={this.props.components}
             data={data}
             icons={this.props.icons}
-            localization={{ ...MTableBody.defaultProps.localization.editRow, ...this.props.localization.editRow }}
+            localization={{ ...MTableBody.defaultProps.localization.editRow, ...this.props.localization.editRow, locale: this.props.localization.locale }}
```
mbrn: It seems you forgotten this.

Issue #676, opened 2019-06-05
Selection and Actions in the same Table Row


RobertLbebber:
Allow tables to have rows with both selection Checkboxes and Action Buttons

mbrn:
Hi @RobertLbebber 

This feature requested by others many times. Firstly i advise them to using custom render feature. But i decided to add this feature asap.

RobertLbebber:
Yea I think its important because it was weird to track down why the actions were disappearing when the code looks all functionally correct

KATT:
I really appreciate your work on Material Table. I've started using it heavily and looking forward to contributing.

Could lend a hand on implementing this, I think.

As I see it there's three types of actions currently:

1. `isFreeAction` top bar action
2. Top bar action _when rows are selected_
3. Actions on rows

The type of #2 and #3 are currently implicitly defined based on if it's a selectable table. Correct me if I'm wrong or missed something

I would propose one or these two approaches (backwards-compatible with just deprecating  the old API):

1. Move row actions to `columns[]`. Would make it very explicit what is where.
2. Explicit `position`-prop in the `actions[]` array with 4 options + deprecate `isFreeAction`
   1. `auto` (default) - existing behaviour (it can figure out the one of the below)
   2. `header`
   3. `headerOnSelect` - Show when a row is selected
   4. `row`

yuri3:
Hi mbrn, can you tell me please the above feature requested by RobertLbebber, will be added?
if so when?

ahameleon:
Also waiting for this feature. Any updates as of yet?

jayarjo:
Related: https://github.com/mbrn/material-table/issues/1107. I've also posted solution to this problem on [SO](https://stackoverflow.com/questions/58150039/material-table-how-to-do-selectable-and-editable-table/58153038#58153038) 

@KATT unfortunately only @mbrn has enough insight on the project to evaluate your suggestion. Hope he will the time to post some thoughts. 

Until then you folks could try the solution from the thread I've referenced above.

jayarjo:
Also related: https://github.com/mbrn/material-table/issues/648

tafelito:
@jayarjo in addition of having actions in the row level, it would be great to also have an option in the toolbar to hide the actions there if they are shown at the row level. 
what I'm doing now is using your example for the TableBodyRow and  the Toolbar but I added an option to not show the actions if it's a row action here

[...]

PR #1156, opened 2019-10-04
Add position field to action

CoolCyberBrain:
## Related Issue
#676 

### Can now show action on row while `option.select === true`
Implements idea 2 of https://github.com/mbrn/material-table/issues/676#issuecomment-502665435
> I really appreciate your work on Material Table. I've started using it heavily and looking forward to contributing.
> 
> Could lend a hand on implementing this, I think.
> 
> As I understand it there's three types of actions currently:
> 
> 1. `isFreeAction` top bar action
> 2. Top bar action _when rows are selected_
> 3. Actions on rows
> 
> The type of 2 and 3 are currently implicitly defined based on if it's a selectable table. (Please correct me if I'm wrong or missed something)
> 
> I would propose one or these two approaches (backwards-compatible with just deprecating the old API):
> 
> 1. Move row actions to `columns[]`. Would make it very explicit what is where.
> 2. Explicit `position`-prop in the `actions[]` array with 4 options + deprecate `isFreeAction`
>    
>    1. `auto` (default) - existing behaviour (it can figure out the one of the below)
>    2. `header`
>    3. `headerOnSelect` - Show when a row is selected
>    4. `row`

Specifically 
> 2. Explicit `position`-prop in the `actions[]` array with 4 options + deprecate `isFreeAction`
>    
>    1. `auto` (default) - existing behaviour (it can figure out the one of the below)
>    2. `header`
>    3. `headerOnSelect` - Show when a row is selected
>    4. `row`

Actions
Toolbar
Row
Filter

### DIFF
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -16,6 +16,13 @@ class MTableAction extends React.Component {
       }
     }
 
+    if (action.action) {
+      action = action.action(this.props.data);
+      if (!action) {
+        return null;
+      }
+    }
+
     if (action.hidden) {
       return null;
     }
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -16,13 +16,6 @@ class MTableAction extends React.Component {
       }
     }
 
-    if (action.action) {
-      action = action.action(this.props.data);
-      if (!action) {
-        return null;
-      }
-    }
-
     if (action.hidden) {
       return null;
     }
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -16,6 +16,13 @@ class MTableAction extends React.Component {
       }
     }
 
+    if (action.action) {
+      action = action.action(this.props.data);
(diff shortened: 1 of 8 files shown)


jayarjo:
Does it takes into account a case when I want action in multiple locations?

CoolCyberBrain:
> Does it takes into account a case when I want action in multiple locations?

You can have multiple actions with everything else the same beside the position

didopop3:
Has this been updated in the website. I don't see any  of the above changes mentioned in https://material-table.com/#/docs/all-props

Light0815:
I also couldn’t find any changes in the docs...
Hence I looked in the code. 
Not sure if this is a bug or a feature but I can not have an action function with (position: "row"). I would need this option in order to disable or hide the actions depending on the contents of the rows

oze4:
I was able to get this to work - [see here for more details](https://github.com/mbrn/material-table/issues/676#issuecomment-629342351)

PR #1157, opened 2019-10-04
Add functional action

CoolCyberBrain:
## Description
After #1156 , I noticed actions that are functions can't indicate their position. This PR adds a new type of action that allows actions to be functions while still indicating its position. This new type of action is of the shape:
```
{
    action: (rowData) => { icon: ..., tooltip: ..., onClick: ... },
    position: ...
}
```

## Related PRs
#1156 

Actions

### DIFF
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -16,6 +16,13 @@ class MTableAction extends React.Component {
       }
     }
 
+    if (action.action) {
+      action = action.action(this.props.data);
+      if (!action) {
+        return null;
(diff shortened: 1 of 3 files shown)


mbrn:
Why you didn't add the position field to Action interface?

mbrn:
I think we don't need this. I solved problem for functional actions with https://github.com/mbrn/material-table/commit/32c400913408fa409e3e2c79c17c6de24273f1b2

We can assume that functional actions are "row" positioned.

PR #1166, opened 2019-10-08
Returns the current data after sorting

Domino987:
## Description
It is [requested](https://stackoverflow.com/questions/58285810/is-it-possible-to-get-the-current-page-data-after-re-ordering-in-material-table) to get the currently displayed data after a sorting happend. This PR returns the current displayed data after onOrderChange.

* Table

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -189,7 +189,7 @@ export default class MaterialTable extends React.Component {
       });
     } else {
       this.setState(this.dataManager.getRenderState(), () => {
-        this.props.onOrderChange && this.props.onOrderChange(newOrderBy, orderDirection);
+        this.props.onOrderChange && this.props.onOrderChange(newOrderBy, orderDirection, this.state.renderData);
       });
     }
   }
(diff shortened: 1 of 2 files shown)


stefansn:
Was this issue fixed?

Domino987:
It is possible to access the data via table ref so if the data is needed, one should always use table ref instead of passing the data around.

eoghanmccarthy:
@Domino987 I posted the tableRef solution on SO but it has been very unreliable and should be avoided. This PR should be considered again.

stefansn:
I also agree that getting the data would be useful. What if you sort the table by two or more columns? Does material-table support multiple sorting?

eoghanmccarthy:
@Domino987 same link you posted. that's me here: https://stackoverflow.com/questions/58285810/is-it-possible-to-get-the-current-page-data-after-re-ordering-in-material-table#comment102945551_58285810

eoghanmccarthy:
What would be great, regardless of whether it's sorted, multiple sorted, filtered, etc, is to have the visible data array returned.

eoghanmccarthy:
@Domino987 Just wondering if there's a way forward for this issue. Is it possible to get the visible list array with the current api (tableRef, etc.)? Or does it need added functionality?

raztiff:
Hate to be that guy, but any movement on this please, makes redux integration a tad awkward (from my understanding thus far). Thanks :)

Issue #1014, opened 2019-08-21
Issues using material-table with redux


whuey1992:
When I use redux to handle the data that is being passed to material-table. I run into this error
"Invariant Violation: A state mutation was detected between dispatches, in the path `user.data.0.tableData`. This may cause incorrect behavior". It seems that material-table is doing some kind of state change when my dispatches are occurring. Only solution I have found so far is creating a clone of the data in redux like this data={JSON.Parse(JSON.Stringify(this.props.data))

v1siuol:
Same issue when I tried to re-render. 
Originally my code: 
```
const { posts } = props;

// ... In render 
data={ posts }
```

I FIX it by 
```
const _posts = props.posts || []
const posts = _posts.map(item => {
  const im = { ...item }
  return im;
});

// ... In render 
data={ posts }
```

xgarrett:
Hello,
I've figured out same issue in our project.

Temporary fixed by deep cloning of `data` before it's passed to `material-table`.

@mbrn could you please take a look on that?

For me it looks like a serious bug that some kind of internal state from library is appended to data being passed to a component.
If `material-table` needs to have its own copy of data with extra params it should do it internally.

PR #1174, opened 2019-10-10
Remove data mutation

Domino987:
## Related Issue
#1014 

## Description
This PR removes the mutation of the provided data by copying the provided data and not writing to the provided props directly. 

This is in line with the best practices since props should never be mutated and this causes unforeseen side effect. If the data comes from redux or another state management library, these changes persist over the lifetime of the session and might even be played back into the backend if the data is updated and the tabledata is passed as well, since the global object is changed.

This also makes it difficult to reset the selection if the same data is used for multiple tables in which case selection in one table effects the selection of the other, even so they should be separated.

Additionally, the mutation is not mentioned in the documentation which makes these changes also unpredictable.

I am aware that this change will reset the selection, the extended panels, edits etc. on data change, but this is in my opinion not a problem since the table should reset itself, if new data is received. And if it is needed to persist the selection, one could use the tableRef to access the data and use that to keep the selection.

Let me know what you think and if you agree with me so we can adress this issue.

* Data-Manager

## Additional Notes
I would really like this change to happen, since it makes the api more predictable and does not mutate my data ;)

### DIFF
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -46,11 +46,14 @@ export default class DataManager {
     this.selectedCount = 0;
 
     this.data = data.map((row, index) => {
-      row.tableData = { ...row.tableData, id: index };
-      if (row.tableData.checked) {
+      const localRow = {
+        ...row,
+        tableData: { ...row.tableData, id: index }
+      }
+      if (localRow.tableData.checked) {
         this.selectedCount++;
       }
-      return row;
+      return localRow;
     });
 
     this.filtered = false;
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -49,7 +49,7 @@ export default class DataManager {
       const localRow = {
         ...row,
         tableData: { ...row.tableData, id: index }
-      }
+      };
       if (localRow.tableData.checked) {
         this.selectedCount++;
       }


Domino987:
Hi @jpmtrabbold , Yes it may be a breaking change for some as I described in the commit description. If you rely on the tableData to persist or something similar it may change. The users should not rely that onRowClick returns the same object. What is your use case, that makes it necessary for it to be that same object? Can you describe it, so that we can work on a solution?

Domino987:
hi @jpmtrabbold , can you tell me about your use case that you need the exact reference to the original data object?

jpmtrabbold:
Let's say I have:
source of truth: 10 rows
filtered data used with material-table: 4 rows (out of the 10 original)

User clicks on that row, and I receive the onRowClick callback. I need to know what's the original item in the source of truth, so I use sourceOfTruthArray.indexOf(onRowClickItem). 

I know that I could solve that just by using ids, but we have too many places using that.

Also, it's expected that the callback should hold the same reference. This week I wrote a component to replace material-table in our use cases, and I found a great solution for both not mutating the props data and also not having separate references for the component (waste of memory), it's working pretty well.

All in all, don't worry about me - if you are going to change it, do it for the sake of having a more predictable component because I'm not depending on it anymore. Keep up the good work!! Cheers bro

Issue #786, opened 2019-07-03
Add a handler `onRowEdit` and `onRowEditCancel`


AntoniusGolly:
For when `editable` objects are provided...

**Is your feature request related to a problem? Please describe.**
I check for errors when validating input data on editing or adding new items. Those errors I want to clear when the editing is canceled. 

**Describe the solution you'd like**
There is already some nice handlers as `onOrderChange` or `onRowClick`. I would like to have an additional handler `onRowEdit` when editing is started and `onRowEditCancel` when the X is pressed in editing mode. 

**Additional context**
Also, I think it makes great sense to have a handler for when the `detailPanel` is opened/closed.

hGica:
I am also searching for a way to do this. I also couldn't manage to successfully override editrow component to get the functionality as there is no documentation on how to do this.

waisbek:
It's a great idea, I also need identify when the user is editing.

hGica:
@waisbek  see my answers on https://github.com/mbrn/material-table/issues/751 though I wasn't able to get access to is editing

CycoPH:
Internally there is the 'onEditingCanceled' event.  It just has to be exposed to the outside.
This would make external messaging way easier if we know when cancel was pressed.

Rombusevil:
I've stumbled upon this issue due to needing functionality to handle a user message when quitting the page with unsaved data. I've made the PR providing a callback that will be called with a boolean indicating a "isEditing" status. 
You can try it out installing it like this  `npm install git://github.com/Rombusevil/material-table.git#onEditingStatusChange-build --save`

PR #1184, opened 2019-10-12
added onEditingStatusChange callback.

Rombusevil:
## Related Issue
#786 Allows to setup a callback for listening the state of the table. True if it's in "dirty" mode (adding, updating, or deleting) and false if it's just displaying data.

## Description
Adds a new prop `onEditingStatusChange` that expects a function with one boolean argument describing if the table is in "edit" mode or not.

## Additional Notes
I want this functionality simply to show a message if the table is in edit mode and the user wants to quit the page.

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -480,6 +480,8 @@ export default class MaterialTable extends React.Component {
 
   render() {
     const props = this.getProps();
+    const hasAnyEditingRow= !!(this.state.lastEditingRow || this.state.showAddRow);
+    !!props.onEditingStatusChange && props.onEditingStatusChange(hasAnyEditingRow);
 
     return (
       <DragDropContext onDragEnd={this.onDragEnd}>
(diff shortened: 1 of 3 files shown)


src/material-table.js
```diff
@@ -480,6 +480,8 @@ export default class MaterialTable extends React.Component {
 
   render() {
     const props = this.getProps();
+    const hasAnyEditingRow= !!(this.state.lastEditingRow || this.state.showAddRow);
+    !!props.onEditingStatusChange && props.onEditingStatusChange(hasAnyEditingRow);
```
mbrn: Calling an event on render function does not make sense? You can handle this change on material-table.js

mbrn:
You can reopen pr after make changes that requested.

mosofsky:
I would still like to have a callback when any row goes into edit state.  I need this for a modal dialog box which contains a MaterialTable.  I need to disable the modal's OK button when any row becomes editable.

The solution at https://stackoverflow.com/questions/58144937/check-if-a-material-table-row-is-still-in-edit-mode#comment134859367_58145627 is inadequate because it's passive.  It waits until the user clicks a button and then has to show an error message.  Instead, we'd like a proactive solution.  A callback like the one proposed in this ticket, namely onEditingStatusChange, seems like it would be great.

However, I want to make sure that the handler for onEditingStatusChange is able to change the React state without causing the MaterialTable to slip from editable state to uneditable state.  I suspect that might be an issue.  So here are som test cases:

Given a modal
  And a MaterialTable in the modal
When the user adds a row
Then the modal's OK button should become disabled
  But the added row should still be in the edit state

Given a modal
  And a MaterialTable in the modal
  And a populated row
When the user edits the row
Then the modal's OK button should become disabled
  But the row being edited should still be in the editable state

PR #1190, opened 2019-10-14
Remove approve from props

Domino987:
## Description
This clears a warning that approve should not be spreaded to Select and textfields. So it gets removed from the spreaded props.

* Edit Field


mbrn:
Same comment with #1145

PR #1197, opened 2019-10-14
Modify double quotation mark for consistency

leesungrin:
## Related Issue
Coding convention.

## Description
All import files should be in single quotation mark.


jayarjo:
No sorry. Not pulling this in. Once author find some spare time, there's PR in the queue that will format and prettify whole code base ;)

PR #1199, opened 2019-10-15
Add row action option

tafelito:
## Related Issue
#676 

## Description
Add the option isRowAction to allow showing actions on a row level. If an action has the property isRowAction set to true, the action won't be shown in the toolbar.

* MTableToolbar
* MTableBodyRow
* MTableBody

## Additional Notes
For some cases, sometimes is better to have individual actions on every row instead of having to select them. For example, an edit action is OK to show per row, but a remove could be selectable. Maybe sometimes is also ok to have both options as well

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -81,8 +81,23 @@ class App extends Component {
                     selection: true,
                     columnsButton: true,
                     filtering: true,
+                    actionsColumnIndex: -1,
                     defaultExpanded: row => row.surname === 'C'
                   }}
+                  actions={[
+                    {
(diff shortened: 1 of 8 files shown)


jayarjo:
There's a related PR here: https://github.com/mbrn/material-table/pull/1156. Would you mind to take a look if it serves the same purpose?

tafelito:
Apparently it does. And I actually like the naming options a bit better there. If everything works I would be inclined to close this one and merge that other one

mbrn:
I think that #1156 is better solution. Thanks for contribution.

Issue #1153, opened 2019-10-03
Incorrect Issue Type Definition for DetailPanel


ririvas:
The interface for DetailPanel appears to have a definition mismatch with the actual implementation.

### Typescript Definition
What's in source 
```
export interface DetailPanel<RowData extends object> {
  disabled?: boolean;
  icon?: string | React.ReactElement<any>;
  openIcon?: string | React.ReactElement<any>;
  tooltip?: string;
  render: (rowData: RowData) => string | React.ReactNode;
}
```
The corresponding code for the definition above
```
<MaterialTable
	...
	detailPanel={[
		rowData => ({
			icon: <Add/> ,
			render: () => ...
		})
	]}
/>
```
**The above code will compile but fail during run time**

### Solution
Change the definition to the following
```
export interface DetailPanel<RowData extends object> {
  disabled?: boolean;
  icon?: string | React.ComponentType<any>;
  openIcon?: string | React.ReactElement<any>;
  tooltip?: string;
  render: (rowData: RowData) => string | React.ReactNode;
}
```
And use the following implementation
```
<MaterialTable
	...
	detailPanel={[
		rowData => ({
			icon: Add ,
			render: () => ...
		})
	]}
/>
```
**The code above works**
I expect the same applies for openIcon

jayarjo:
PRs are welcome!

ririvas:
Sounds good.

ririvas:
This has been merged

PR #1201, opened 2019-10-15
Fixed typescript definition for DetailPanel interface to use correct …

ririvas:
…Component type

## Related Issue
#1153 

## Description
Update to the type definition file (index.d.ts) to correct the interface for DetailPanel. The existing definition mismatched with actual js implementation. 

Also fixed comma's in the definition to semi-colons. 

This should only affect the typescript definition.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -62,8 +62,8 @@ export interface QueryResult<RowData extends object> {
 
 export interface DetailPanel<RowData extends object> {
   disabled?: boolean;
-  icon?: string | React.ReactElement<any>;
-  openIcon?: string | React.ReactElement<any>;
+  icon?: string | React.ComponentType<any>;
+  openIcon?: string | React.ComponentType<any>;
   tooltip?: string;
   render: (rowData: RowData) => string | React.ReactNode;
 }
@@ -80,19 +80,19 @@ export interface Action<RowData extends object> {
 
 export interface EditComponentProps<RowData extends object> {
   rowData: RowData;
-  value: any,
-  onChange: (newValue: any) => void,
-  columnDef: EditCellColumnDef,
+  value: any;
+  onChange: (newValue: any) => void;
+  columnDef: EditCellColumnDef;
 }
 
 export interface EditCellColumnDef {
-  field: string,
-  title: string,
+  field: string;
+  title: string;
   tableData: {
-    filterValue: any,
-    groupOrder: any,
-    groupSort: string,
-    id: number,
+    filterValue: any;
+    groupOrder: any;
+    groupSort: string;
+    id: number;
   }
 }
 


Domino987:
Can you show us the exact implementation and how it fails for you? I cannot seem to be able to reproduce it.

ririvas:
@Domino987 

Here is a code sandbox [link](https://codesandbox.io/s/material-table-demo-z1jek)
```typescript
// this compiles but errors out during runtime
icon: <div />
// this does not compile but would work
icon: () => <div />
```

jayarjo:
@ririvas 

> ```ts
> // this compiles but errors out during runtime
> icon: <div />
> // this does not compile but would work
> icon: () => <div />
> ```

Does your solution make both work? I'm not exactly proficient in typings. From what I see on the `React.ComponentType<Props>` [description](https://flow.org/en/docs/react/types/#toc-react-componenttype) on Flow site though:

> `React.ComponentType<Props>` does not include intrinsic JSX element types like div or span. See `React.ElementType` if you also want to include JSX intrinsics.

I think typings that you suggest are also not fully inclusive.

ririvas:
@jayarjo  `React.ElementType`  is an extension of `React.ComponentType`. I think that will also work. However, as I understand, the purpose of using that type would be to understand native html components, i.e. replace `() => <div/>` with `div`. And I haven't been able to pass `div` as a value. Regardless of the type definition, `<Component/>` is not a valid input as that "compiles". 

See source below.
```typescript
const CustomIcon = ({ icon, style }) => typeof icon === "string" ? <Icon style={style}>{icon}</Icon> : React.createElement(icon, { style });
```

Happy to replace `React.ComponentType` with `React.ElementType`, both are fine

Domino987:
@jayarjo , @ririvaa , Since the <div/> does not work with the table at the moment and it should be a function we might change it to `string | () => React Element<any>`. What do you say ?

ririvas:
I think `string | React.Element<any>` would still be invalid. `string | React.ElementType<any>` or `string | React.ComponentType<any>` would work, however the `string | ` is unnecessary. 

Below is valid for either ComponentType or ElementType
```typescript
detailPanel={[
          rowData => ({
            icon: "add",
            render: () => <div>test panel</div>
          })
        ]}
```
```typescript
detailPanel={[
          rowData => ({
            icon: () => <div/>, // custom icon def
            render: () => <div>test panel</div>
          })
        ]}
```

mbrn:
@Domino987 , @ririvas 

What do you think about React.ReactType<Props>?

Domino987:
@mbrn since ReactType is deprecated, we should use ElementType from the official typings:
declare namespace React { 

// // React Elements // ----------------------------------------------------------------------

 type ElementType<P = any> = { [K in keyof JSX.IntrinsicElements]: P extends JSX.IntrinsicElements[K] ? K : never }[keyof JSX.IntrinsicElements] | ComponentType<P>; 


/** * @deprecated Please use `ElementType` */ type ReactType<P = any> = ElementType<P>; 


type ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>; type JSXElementConstructor<P> = | ((props: P) => ReactElement | null) | (new (props: P) => Component<P, any>);

ririvas:
@Domino987 want me to just close this pull request? I could also submit one for React.ElementType. Current request is to switch detail panel to ComponentType

Domino987:
No you should just change it to Element type. The problem with Component Type is that it disallows strings. Let me think about it.

PR #1205, opened 2019-10-15
Hide filter icon per column via column property hideFilterIcon:true

ryanintulsa:
## Related Issue
none

## Description
This is a cosmetic change that allows completely hiding the Filter icon on a per-column basis.

## Related PRs
none

* this is an optional additional column property that won't affect existing code, as the default value is false, which shows the Filter icon


mbrn:
Why you don't add a general hideFilterIcon to options. I don't mean remove it from columns. But there should an option to hide all filter icons.

What do you think?

ryanintulsa:
> Why you don't add a general hideFilterIcon to options. I don't mean remove it from columns. But there should an option to hide all filter icons.
> 
> What do you think?

Good idea. I added hideFilterIcons (plural to distinguish a bit from the column option).

ryanintulsa:
Not sure why it's showing some formatting changes. I don't recall making those. Let me know if you need me to fix that or if it's alright as is.

ryanintulsa:
@mbrn this PR is ready for re-review with the requested changes. Thanks.

PR #1206, opened 2019-10-15
Hide filter row programmatically

ryanintulsa:
## Related Issue
none

## Description
A cosmetic change to allow programmatically hiding the filter row via new table option hideFilterRow:true. Note that by default hiding the filter row clears the filters. This behavior is desirable so that there is not hidden data without indication of a filter applied.

## Related PRs
none

* This is an additional optional table option which defaults to false, so existing code will not be affected.


mbrn:
Hi @ryanintulsa 

Thanks for contribution. But this pr does not make sense. You can use options.filtering. To save filtering data before hiding filter row, We can find a better solution.

Issue #1037, opened 2019-08-26
typescript error in intrinsic types for columns 


tafelito:
setting columns field from a state variable throws a type error if is not declared in line

```js
const [state, setState] = React.useState({
    columns: [
      { title: 'Name', field: 'name' },
      { title: 'Surname', field: 'surname' },
      { title: 'Birth Year', field: 'birthYear', type: 'numeric' },
      {
        title: 'Birth Place',
        field: 'birthCity',
        lookup: { 34: 'İstanbul', 63: 'Şanlıurfa' },
      },
    ],
    data: [
      { name: 'Mehmet', surname: 'Baran', birthYear: 1987, birthCity: 63 },
      {
        name: 'Zerya Betül',
        surname: 'Baran',
        birthYear: 2017,
        birthCity: 34,
      },
    ],
  });

  return (
    <MaterialTable
      title="Disable Field Editable Preview"
      columns={state.columns}
      data={state.data}
      ....
   )
```

this throws this error

`
The expected type comes from property 'columns' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<MaterialTable<{ name: string; surname: string; birthYear: number; birthCity: number; }>> & Readonly<MaterialTableProps<{ name: string; surname: string; birthYear: number; birthCity: number; }>> & Readonly<...>'
`

If instead I do this

```js
 <MaterialTable
      title="Disable Field Editable Preview"
      columns={[
        { title: 'Adı', field: 'name' },
        { title: 'Soyadı', field: 'surname' },
        { title: 'Doğum Yılı', field: 'birthYear', type: 'numeric' },
        {
          title: 'Doğum Yeri',
          field: 'birthCity',
          lookup: { 34: 'İstanbul', 63: 'Şanlıurfa' },
        },
      ]}
      data={[
        { name: 'Mehmet', surname: 'Baran', birthYear: 1987, birthCity: 63 },
      ]}
     />
```

this will not throw any error, and the types are correct

baomastr:
also having ts error about IntrinsicAttributes:

`TS2322: Type '{ children: Element; value: string; rowData: Task; onChange: (newValue: any) => void; columnDef: EditCellColumnDef; }' is not assignable to type 'IntrinsicAttributes'.
  Property 'children' does not exist on type 'IntrinsicAttributes'.`

jayarjo:
PRs are welcome and very much encouraged!

DustinRobison:
[...]

PR #1212, opened 2019-10-16
Updated demo.js to use hooks

DustinRobison:
## Related Issue
https://github.com/mbrn/material-table/issues/1037

## Description
I updated the `demo/demo.js` to be a modern React functional component and use hooks `React.useState` as I was testing the related issue and thought it may be a nice thing to share back to material-table

* Only development code, `demo/demo.js` file

## Additional Notes
The above issue (https://github.com/mbrn/material-table/issues/1037)  can be closed because this seems to work without any errors

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -31,42 +31,40 @@ for (let i = 0; i < 1; i++) {
   bigData.push(d);
 }
 
-class App extends Component {
-  tableRef = React.createRef();
+const App = () => {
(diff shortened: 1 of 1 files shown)


mbrn:
Hi @DustinRobison 

Thanks for contribution. Now, we don't need hooks. We should first focus to issues. Please feel comfort to send more prs :)

Issue #932, opened 2019-08-05
Sort columns by drag & drop with hidden column


juldrixx:
**Describe the bug**
When one column has the hidden feature to true the feature to order columns by dragging them doesn't work properly.

Tibblist:
I am also experiencing this issue when using the columnsButton option and toggling 1 or more columns to be hidden

mieber:
The same here. A fix would be highly appreciated.

ryanintulsa:
> I am also experiencing this issue when using the columnsButton option and toggling 1 or more columns to be hidden

@Tibblist I saw this issue as well, but I think it is unrelated to this drag & drop bug. Please create another issue. I'll take a look at this one and then that one if someone doesn't beat me to it, because both of these are must fix issues for my project.

ryanintulsa:
My PR to fix this is here: https://github.com/mbrn/material-table/pull/1228

PR #1228, opened 2019-10-18
#932: fixed column drag and drop when some columns are hidden

ryanintulsa:
## Related Issue
https://github.com/mbrn/material-table/issues/932

## Description
Drag and drop column reordering was not considering hidden columns.

## Related PRs
none

data-manager


.gitignore
```diff
@@ -1,6 +1,5 @@
 .docz
 .vscode
-dist
```
mbrn: Why did you removed this line? This makes your pr unreadable.

ryanintulsa:
@mbrn sorry, an oversight. Please look again. If you're curious why this happens, it's because when you aren't ready to publish a fork to npm you can still reference it in github directly in package.json like "material-table": "git://github.com/ryanintulsa/material-table.git#myBranchName", but for that to work you must check in dist. It's confusing hopping back and forth, and I wish git had a better solution.

ryanintulsa:
I've got a boundary condition error. I'm working on the fix.

austinterris:
This is still an issue on 1.56.1.  It seems to not work the first time you try dragging over a hidden column but if you do the same action again then it will work.

PR #1232, opened 2019-10-19
add multiline textfield support

flarpy:
Using textfields with anything more than a small amount of data is unworkable in edit mode. This allows  user to specify using a textarea field


mbrn:
Hi @flarpy 

Thanks for contribution. But type field means type of data. It does not represent the component type. 

You can use editComponent of column to make textField multiline:

```javascript
import {MTableEditField} from 'material-table'
---
{ title: 'Adı', field: 'name', editComponent: (props) => <MTableEditField {...props} multiline/> },
```

hatchli:
Just wanted to express my thanks for this! Thought I had to do some weird css stuff, but nope! A simple, elegant solution :)

Issue #1233, opened 2019-10-20
Add a minBodyHeight option (to go with the maxBodyHeight option)


mfreeborn:
I'm using material-table in my web app and I'm really liking it. However, I've come across a use case for having access to a minBodyHeight option, to go along with the maxBodyHeight option already available.

I have demonstrated my case pictorially [here](https://imgur.com/a/i1DS7SW).

I've forked the repo and committed the very few changes necessary to implement this, and I can submit the PR imminently for review.

I'm not sure how I would update the docs on material-table.com, but I'd like to do that too, if possible.

PR #1234, opened 2019-10-20
#1233 Add minBodyHeight to available options

mfreeborn:
As per issue #1233


mbrn:
Thanks for contribution. You can fork mbrn/material-table.com to update docs. And send a pr.

PR #1242, opened 2019-10-22
display length of grouped data

mbwatson:
## Description
I often would like to know the number of items upon grouping data by a column. This adds that information to the grouped data's title row.


martin-pepgit:
This is a nice feature, but imho. this should be optional.
Current patch enforces this feature for everyone. I donÄt think this is a good idea.

mbwatson:
Yes, you're right. That was presumptuous of me. I'll work on making this configurable.

mbwatson:
i'm going about this in a slightly different way now, so i think it's best to close this PR in lieu of an improved solution.

Issue #1235, opened 2019-10-20
When using function for Action, re-renders fully add/remove custom component


rylandg:
**Describe the bug**

When adding an action implemented with a function (as opposed to an icon), the DOM seemingly fully removes and re-inserts the action element for each re-render. This obviously effects performance in an unacceptable way, causing all rows to briefly flash on any search or just plain loading. 

**To Reproduce**

1. Create a table with actions
2. Add action that uses a custom component

```jsx
          (rowData: any): any => ({
            icon: () => <a href='google.com'><img src={img} /> </a>
            tooltip: 'Google link',
            onClick: (event: any, rowData: any) => {}
          }),
```

3. Search in Material table
4. See flickering of the image as it's continuously removed from the DOM

**Expected behavior**

The action only should re-render if the component changes. If you provide a function that returns an icon string (not a custom component), things behave correctly:

```jsx
          (rowData: any): any => {
              icon: 'fiber_new',
              tooltip: 'New updates!',
              iconProps: { color: 'secondary' },
              onClick: (event: any, rowData: any) => {},
            }
```

If I had to guess, it's something about this conditional in `m-table-action.js`

```jsx
          {typeof action.icon === "string" ? (
            <Icon {...action.iconProps}>{action.icon}</Icon>
          ) : (
              <action.icon
                {...action.iconProps}
                disabled={action.disabled}
              />
            )
          }
```

kwiatkk1:
I've managed to reproduce and fix this. Looks like calling render prop directly as opposed to via jsx wrapper, causes no re-render.

rylandg:
@kwiatkk1 awesome work. Thanks for figuring  this out, already implemented your fix in my fork!

PR #1250, opened 2019-10-24
#1235: prevent actions icons re-render

kwiatkk1:
## Related Issue
#1235

## Description
Prevent action icons to re-render unconditionally.

* MTableAction

### DIFF
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -37,10 +37,10 @@ class MTableAction extends React.Component {
           {typeof action.icon === "string" ? (
             <Icon {...action.iconProps}>{action.icon}</Icon>
           ) : (
-              <action.icon
-                {...action.iconProps}
-                disabled={action.disabled}
-              />
+              action.icon({
+                ...action.iconProps,
+                disabled: action.disabled
+              })
             )
           }
         </IconButton>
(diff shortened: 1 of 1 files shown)


kwiatkk1:
@mbrn what do you think about @TheGoodSpinach suggestion to make some additional changes within this pull-request? Or do you rather want to keep changes small and to open separate one for it?

mbrn:
> @mbrn what do you think about @TheGoodSpinach suggestion to make some additional changes within this pull-request? Or do you rather want to keep changes small and to open separate one for it?

I prefer small bites :)

mbrn:
Hi @kwiatkk1,

 I tested this pr. It is not working. Please check it again. Especially with editable feature.

kwiatkk1:
Hey @mbrn, thanks for review! You're right!

I've tested only a case in which actions were added manually and they were string/function as described in https://material-table.com/#/docs/features/actions.

But when the `editable` option is used, it adds actions that have `icon` property that is a product of React.createRef. And that is not a string nor function but a Component. 

So I added additional check in a code to handle this. Another option would be to change the way default props are defined, but i assume that it's a way more work.

kwiatkk1:
Thanks for accepting this.

Now time for the issue pointed out by @TheGoodSpinach. I'll deal with it in a separate PR.

Issue #1223, opened 2019-10-18
defaultSorting on column still active when sorting = false


frankforpresident:
**Describe the bug**
When you provide sorting: false to options, defaultSort: 'desc' from a column is still applied. When reading the documentation this should not be the case.

"To make sorting available you must open sorting in options." and "sorting | boolean | true | Flag to activate or disable sorting feature of table"

**To Reproduce**
Install 1.52.0
render a default table with options={{ sorting: false }}
provide a column with defaultSort

**Expected behavior**
defaultSort is not applied on the data

**Actual behavior**
defaultSort is sorting the data.

sshanzel:
When you apply `sorting: false` also on `column` props with `defaultSort`, it happens. Though this is fixed on the referenced PR

sshanzel:
This has been fixed with the merging of referenced PR. Kindly close the issue if you are now satisfied with the fix. Thanks a lot for the report!

mbrn:
> This has been fixed with the merging of referenced PR. Kindly close the issue if you are now satisfied with the fix. Thanks a lot for the report!

We can close the issue

PR #1258, opened 2019-10-27
bugfix: disable defaultSorting when sorting = false

sshanzel:
## Related Issue
#1223 

## Description
Config for validating when applying `defaultSort` is not checking the options for `sorting: false`. Also when applying it directly at the `column`. As per the documentation, this should not be the case.

Sorting of columns.

## Additional Notes
Just let me know if you have any concerns so I can immediately coordinate.

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -55,8 +55,8 @@ export default class MaterialTable extends React.Component {
   setDataManagerFields(props, isInit) {
     let defaultSortColumnIndex = -1;
     let defaultSortDirection = '';
-    if (props) {
-      defaultSortColumnIndex = props.columns.findIndex(a => a.defaultSort);
+    if (props.options.sorting !== false) {
+      defaultSortColumnIndex = props.columns.findIndex(a => a.defaultSort && a.sorting !== false);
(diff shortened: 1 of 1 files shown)


src/material-table.js
```diff
@@ -55,8 +55,8 @@ export default class MaterialTable extends React.Component {
   setDataManagerFields(props, isInit) {
     let defaultSortColumnIndex = -1;
     let defaultSortDirection = '';
-    if (props) {
-      defaultSortColumnIndex = props.columns.findIndex(a => a.defaultSort);
+    if (props.options.sorting !== false) {
```
mbrn: You may check for props is not undefined. // (props && props.options.sorting !== false)

sshanzel: Though i've tested passing nothing on the props and it prompted nothing, yes i'd apply now safe checking. thanks a lot for the review!

sshanzel:
Good day! I've now applied safe checking for `props` when applying `defaultSort`. Thanks a lot for the review!

sshanzel:
You're welcome!

Issue #1213, opened 2019-10-17
CellStyle is not actually working


sshanzel:
**Describe the bug**
Data placed in `options` for `cellStyle` is not actually being passed down to the cells.

**To Reproduce**
Apply any styling property to `cellStyle`. It won't reflect in any case. 

**Expected behavior**
Cell style should follow as the option provided

sshanzel:
I already checked and forked the project. Will make a PR once someone declares this as a valid issue

jayarjo:
@sshanzel please do.

sshanzel:
@jayarjo i've actually created the PR (#1214) already but this issue wasn't verified yet back then, that's why it was closed for the meantime. Should i be creating another PR or we would reopen the referenced PR?

Issue #552, opened 2019-05-08
Change Header row Height


denis-sk:
Make it possible to change the header row height.

Now I do not know how to do it. I can change only header cell style.

mbrn:
Hi @denis-sk 
 You can use options.headerStyle to change header rowHeight. 

![image](https://user-images.githubusercontent.com/7895451/57427077-f8361e80-722a-11e9-8cc0-6e47ab975e2c.png)

denis-sk:
No. Header Style change th cell style, but tr style not changes. 

![image](https://user-images.githubusercontent.com/26207133/57521379-cb683100-7328-11e9-9335-a6c9355a2631.png)

mbrn:
I will add cellStyle to options to effect all cells.

denis-sk:
Do you mean the cellstyle will be applied to the header cell and row cell? If yes, I think they should work independently of each other.

mbrn:
No it will be applied to all cells(not headers)

sshanzel:
#1214 PR for passing cellStyle down to ColumnDef for options to take effect.

sshaw:
Problem here is `tr` has a height of `56px` so if cell height is smaller is does not matter. `56px` is used.

EdmarPereira:
I found a solution for my use, I need to edit this property to change the size of the table header and table body:
.MuiTableCell-root {
  height: 10px !important;
  padding: 0px !important;
  font-size: 0.7rem !important;
}

.MuiButtonBase-root{
  height: 10px !important;
}

with this you can adjust for you

Forsaiken:
I found a solution changing the paddingTop and paddingBottom on HeaderStyle

asok-mirror:
> I found a solution for my use, I need to edit this property to change the size of the table header and table body:
> .MuiTableCell-root {
> height: 10px !important;
> padding: 0px !important;
> font-size: 0.7rem !important;
> }
> 
> .MuiButtonBase-root{
> height: 10px !important;
> }
> 
> with this you can adjust for you

This seems to be hiding the footer as well

asok-mirror:
The row height is not getting applied though I can see the height property at tr level. weird though!! 

Any thoughts?

solisadeqi:
height dose not work :( , have tried multiple solutions

yashaswinibprasad87:
any fix for the issue?

jason-zeiler-bayer:
Could use some help here as well if anyone has found a solution?

PR #1268, opened 2019-10-29
Bugfix: Passing cellStyle to ColumnDef props

sshanzel:
## Related Issue
#1213 

## Description
`options.cellStyle` props is not actually being passed down to `cell` `columnDef` props but already being used. As a result, styles don't reflect.

Custom styling of Cells as mentioned in #552 doesn't work.

## Additional Notes
If you have any suggestions or concerns i am more than happy to hear it. Thanks a lot for this tool, i am one of those who's actually using it! :D


sshanzel:
You're welcome! There's more to come!

Issue #1265, opened 2019-10-29
The table is fully rerendered on parent state update since 1.53.0


sebastiendan:
**Describe the bug**
I updated to 1.53.0 and noticed the table was fully rerendered when the parent component state was updated. This was a big deal in my situation: I have an EventSource that receives data pushed by the server and renders new data in new rows. Because of this new behaviour each data push (followed by a state update in the parent) led the table to get back to init configuration (first page, default page size, etc.).

Is this somehow the intended behaviour?

**To Reproduce**
1. Have a paginated table with multiple pages of rows
2. Update the parent state with new values (e. g. update a boolean to its opposite value on an interval)
3. Change page or page size
4. See the table getting back to its init configuration on rerender

**Expected behavior**
Before 1.53.0, you can navigate through the pages and see new rows appear. You can also change any table UI setting without loosing it,

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: 78.0.3904.70 (Official Build) (64-bit)

TJHdev:
I'm having the same issue. New array passed into data is reseting pagination and the search field.

nelstrom:
We're having this issue too. As a temporary fix, downgrading to version 1.52.0 solved the problem for us.

I'm looking at the [release notes for v1.53.0](https://github.com/mbrn/material-table/releases/tag/v1.53.0) trying to figure out which change introduced this behaviour.

nelstrom:
I've found what *might* be a possible cause. 

In `src/material-table.js`, the `setDataManagerFields()` function now includes this line:

    this.dataManager.changeSearchText(props.options.searchText);

The `changeSearchText()` function looks like this:

```
  changeSearchText(searchText) {
    this.searchText = searchText;
    this.searched = false;
    this.currentPage = 0;
  }
```

I haven't had a chance to dig deep into this and figure out how often the `changeSearchText()` function is being called, but I thought the fact that it's setting `this.currentPage = 0` might be a possible cause for this issue.

re-anderson:
@nelstrom I came to the same conclusion. However, I hadn't dug deep enough to see that paging was being reset as well. My (probably duplicate) issue is #1272

jlucier:
I'm now observing better behavior here, things like search text and current page are preserved on re-render, but the chosen "pageSize" and an columns selected/deselected using "columnsButton" are reset to the initial state.

I'm on `1.54.1`.

AndyWorral:
[...]

PR #1225, opened 2019-10-18
Added searchText to options.

naum-d:
Added searchText option to initialize search field before table was render.

PR #1298, opened 2019-11-06
Fixed bug connected with searchText and state.

naum-d:
## Related Issue
#1225  #1265


mlnarik:
type definition for TypeScript needs to be updated, searchText is missing in Options type

PR #1302, opened 2019-11-07
Feature/customize group titles

mbwatson:
## Related Issue
This began as a solution to render the number of items within each group of grouped data. To improve on this, one can supply a function that renders a custom element as a group title.

## Description
I've added the ability to render the result of a function of the grouped data as the contents of the grouped row component. We first set the group row title to the column title. Then we check if a function is supplied via `options.groupTitle`. If so, evaluate it, and use the result as the grouped row title.

For example, one may use the following to render group row titles with a little more information, such as 
**Group: Registered (15)**, which can be accomplished like the following example.

```
<MaterialTable
  tableRef={this.tableRef}
  columns={columns}
  data={data}
  title="Table Title"
  options={{
    grouping: true,
    groupTitle: group => `Group: ${group.value} (${group.data.length})`,
  }}
/>
```

## Related PRs
N/A

This affects the group row component.

## Additional Notes
I was a little unsure about the inner workings of some of this when I started, but I think I've managed to remove all superfluous stuff I added as I was poking around. I'm still a little unsure about some parts, and I welcome any suggestions on getting this up to par with the usual PRs to this code base.


src/components/m-table-group-row.js
```diff
@@ -107,15 +107,18 @@ export default class MTableGroupRow extends React.Component {
     if (column.lookup) {
       value = column.lookup[value];
     }
-
+    
     let title = column.title;
+    if (typeof this.props.options.groupTitle === "function") {
+      title = this.props.options.groupTitle(this.props.groupData);
+    }
     if (typeof title !== "string") {
```
mbrn: This will be reassinged if title is not string. Please check your code again.

mbwatson:
I think I've addressed the string/non-string concern. I've included two examples in the demo that use string and non-string group titles.

Additionally, the first example illustrates adding the count to the group title that accounts for subgroups if at most two groupings are used.
Screenshot from 2019-11-08 13-46-32
A more robust recursion would be better here, but this just illustrates the possibility for this dynamic title.

GersonDias:
Looks like the index.d.ts file was not updated with this groupTitle option. Also I didn't found this option on the documentation...

PR #1315, opened 2019-11-11
Bugfix: marginTop in TableFilterRow

FastStim:
## Related Issue
Local bug

## Description
Wrong top margin in FilterRow property lookup

Custom style. component Select, style marginTop in MTableFilterRow


ingvaldlorentzen:
Hi, 

I'm wondering why this bugfix was made. It seems to have introduced a bug for me. 
What bug does it actually fix?

It messes up the alignment of the placeholder text when using lookup filtering. Se picture. 
I've been unable to fix this other than downgrading to the last version.

Any suggestions?

image

Issue #1318, opened 2019-11-12
I want to be able to give a custom function to the component know how to get a value from a column when it want to export data to csv


iagolaguna:
I have a table that the values are in a nested array, so I need to map this array to values, for this I used the `render` function that can be added on the `columns` property and so I get the values from my nested array and show on the table, but for export with CSV I can't do that, so when the component tries to export to CSV the values, the component doesn't know how to get the values so my CSV is exported empty, I would like to be able to give a custom function as the `render` that the `columns` property has for when the component tries to export my data to CSV it can know how to get the value from the column, with this i can export complex objects that have fields more deeper

I think that the implementation can be something like :

```
const getDataByColumnName = (columnName,rowData) =>  // implement how to get the value of this specific column
```

the solution is simple if this function isn't undefined when the component will export to CSV, it will use this function instead to access directly the value of the raw data.

Maybe I can help with the implementation of this feature

Creating this feature a think will decrease the necessity of implement all the way to export CSV and others will be able to export the data from the table more easier

iagolaguna:
#1319

PR #1319, opened 2019-11-12
Adding new function to be able to give control to get the value when is exporting csv using the default function

iagolaguna:
## Related Issue
#1318

## Description
Added new function to  `getFieldValueCsv` to get the value only inside `defaultExportCsv`,
if `getFieldValueCsv` be `undefined`, will be used the default function `getFieldValue`.
more details can be read in the issue #1318 

I'm waiting your feedback :)


iagolaguna:
it's missing the type of the new function but, I have a question, in `Options` we will add the dynamic type that extends from RowData or we will move the new function to other property that already use the dynamic type ?

mvitkin:
If I understand correctly, this feature is meant to control how a column is exported to csv, in the case that a column needs to be exported in a different format than it is rendered in the table? I believe this prop should go into the column definition, so it can be applied to a specific column.

iagolaguna:
Yes, you understand well, I get your point and agree with you, I'm going to make those changes now

iagolaguna:
@mvitkin  sorry for the late I was without time to proceed with this feature, but now I'm here, so  I made some changes, now I think it's more in the way that you suggested, what are you think, the code needs other change?

types/index.d.ts
```diff
@@ -127,6 +127,7 @@ export interface Column<RowData extends object> {
   editable?: ('always' | 'onUpdate' | 'onAdd' | 'never' | ((columnDef: Column<RowData>, rowData: RowData) => boolean));
   removable?: boolean;
   render?: (data: RowData, type: ('row' | 'group')) => any;
+  getFieldValue:(rowData: RowData, columnDef: Column<RowData>) => boolean
```
mbrn: Actually, return type is not boolean. It is commonly string.

src/prop-types.js
```diff
@@ -45,6 +45,7 @@ export const propTypes = {
     editable: PropTypes.oneOfType([PropTypes.func, PropTypes.oneOf(['always', 'onUpdate', 'onAdd', 'never'])]),
     removable: PropTypes.bool,
     render: PropTypes.func,
+    getFieldValue: PropTypes.func,
```
mbrn: Can you change name of function as getExportValue?

mbrn:
Hi @iagolaguna ,

Sorry for late response. But can you make some changes, please?

iagolaguna:
@mbrn  thank you for looking this PR, I will make those changes and return to you

src/prop-types.js
mbrn: I think that you didn't see this comment.

mbrn:
It seems that you didn't apply changes that i want. You can reopen this after complete changes.

Issue #1334, opened 2019-11-14
Search slow performance


Kraviecc:
**Describe the bug**
When there is a page size of ~50 (or more) records, searching is very slow (the whole table rerenders when searchText state is changed)

**To Reproduce**
1. Prepare ~30 rows of data
2. Set pageSize to 50
3. Start typing in search box
4. Lags appear

**Expected behavior**
When a user is typing the search phrase, only the toolbar should rerender.

MatiasManevi:
I'll add to this.

Since there is no backend pagination built in. Is it possible to implement virtualization for the not shown rows?

Having a table with 1.000.000 freezes the browser!

PR #1335, opened 2019-11-14
Search optimization fixes

Kraviecc:
## Related Issue
#1334

## Description
Moved searchText state to Toolbar component (still executing onSearchChangeDebounce in parent).

* Toolbar,
* Search


mbrn:
I tested both version for search performance. It seems same. And you didn't changed anything about search algorithm. can you explain the idea please?

Kraviecc:
It's all about how many components will rerender. 

Instead of rerendering whole table when searching, it rerenders only necessary components. It has huge performance impact (especially in my tests) when there are more than 50 rows per page - I had few secs lag when searching. It has no visible influence on less than 50 rows per page.

mbrn:
It is merged. Thanks @Kraviecc

Issue #1222, opened 2019-10-18
showFirstLastPageButtons doesn't work for Stepped Pagination ??


kwasik-com:

jayarjo:
Hm... doesn't it, have an example?

kwasik-com:
`class Test extends React.Component {
  render() {
    return (
      <MaterialTable
        title="Test"
        columns={[
          { title: 'Name', field: 'name' },
          { title: 'Surname', field: 'surname' },
          { title: 'Birth Year', field: 'birthYear', type: 'numeric' },
          {
            title: 'Birth Place',
            field: 'birthCity',
            lookup: { 34: 'İstanbul', 63: 'Şanlıurfa' },
          },
        ]}
        data={[
          { name: 'Mehmet', surname: 'Baran', birthYear: 1987, birthCity: 63 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
          { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
        ]}        
        options={{
          search: false,
          paginationType: "stepped",
          showFirstLastPage: true
        }}
      />
    )
  }
}`

[...]

PR #1342, opened 2019-11-15
Bugfix & Enhancement: Simplified rendering for 'normal' or 'stepped' pagination

sshanzel:
## Related Issue
#1222 

## Description
Simplified the approach for rendering pagination, and fixed the related issue at the same time.

For `stepped` pagination, the first and last page button is not applied in the code.
*

## Additional Notes
I've deleted the file for stepped pagination, instead, i simplified the approach for different paginations. This will help in the future. If you have any concern please let me know.


mbrn:
We might choose this type of pagination options. But i can not merge this to support backward compatibility. Thanks for your effort @sshanzel

sshanzel:
Mind to let me know what can be the other way around? Which part it would be conflicting for backwards compatibility? Can you enlighten me?

Issue #1371, opened 2019-11-21
Material-Table Mutates Item Passed Through data Prop


kbi-daniel:
**Describe the bug**
We are passing an array of objects from our app's Redux state to the data prop. Material Table mutates that Redux state by adding a `tableData` property to each object in the array. This unexpected mutation of our apps Redux state caused some of our data to be corrupted when we saved the state back to our database (now with the inclusion of the `tableData` property.

**To Reproduce**
1. Pass an array of objects into the data prop.
2. console.log the array of objects on component render
3. note that each object in array now has the added `tableData` property.

**Expected behavior**
Material Table should not mutate items passed to data prop.

kbi-daniel:
I created fork that solves the mutation problem.  However, I am not sure if this approach has any other unforeseen consequences.
https://github.com/kbi-daniel/material-table/tree/bug/prevent-data-prop-mutate

Domino987:
This already known was tried to be solved #1174 here.
But this removes the reference to the existing data item on callbacks.
If the object is searched via shallow compares like this:
`onRowClick={(event, row)=> data.find(d => d === r}`
it will fail because the reference changed.

The way material-table saves its data has to be changed to turn the dependencies around like
` row = {id: index, rowData: row }};`
And what is written in tableData is just saved in the row object while the original data stays the the rowData object (same reference and no mutation).
But since accessing the data would have to be changed at all places, this will be a lot of work since tableData has 200 references in the project which all have to be changed to fit the new scheme.
![image](https://user-images.githubusercontent.com/17567991/69364198-30b2bc80-0c92-11ea-9fc1-58037ba1c4de.png)
I am a strong advocate to remove this prop mutation but as I said, it will be a lot of work.

Maybe be can create a branch and work on it together?
@mbrn , what do you think?

amickael:
I've had the same issue, it becomes especially bad when dealing with Redux. Hopefully it will be addressed in the future (I'd be happy to help), but right now I've worked around it by passing data as such:

```js
<MaterialTable data={tableData.map(item => Object.assign({}, item))} ... />
```

That way the data passed to the component is a copy instead of the original, where `tableData` is being supplied by state/prop/selector so it can update with changes.

kbi-daniel:
@amickael we came up with a similar solution as well.

That being said, does it make sense to anyone else (and do you think it would work), to add a small utility function to the package to basically duplicate the code suggested by @amickael.

For example...

```
[...]

PR #1372, opened 2019-11-21
Set row equal to a new object, with new ref, to prevent Material-Tabl…

kbi-daniel:
…e from mutating variable passed through the data prop;

## Related Issue
Relate the Github issue with this PR using #1371

## Description
We are passing an array of objects from our app's Redux state to the data prop. Material Table mutates that Redux state by adding a tableData property to each object in the array. This unexpected mutation of our apps Redux state caused some of our data to be corrupted when we saved the state back to our database (now with the inclusion of the tableData property.

This pull request prevents the mutation of data described above.

### DIFF
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -47,7 +47,7 @@ export default class DataManager {
     this.selectedCount = 0;
 
     this.data = data.map((row, index) => {
-      row.tableData = { ...row.tableData, id: index };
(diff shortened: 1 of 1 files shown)


mbrn:
Hi @kbi-daniel 

Thanks for simple solution. But i am not sure merge this pr. Because some developers used mutated data on their codes. So this might be a problem. 

i can use this improvement later on version 2 as a Major change(breaking change). 

Thanks for your effort.

Issue #743, opened 2019-06-21
Invoking action on any row causes collapsing of nested rows


JKrymarys:
**Describe the bug**
Table used on my website consist with 3 layers of nested rows. Each of them has same set of actions. The issue is that, when I am invoking actions (clicking the button) on the any of the nested rows it causes collapse of all. 

This is unwanted, as it breaks the functionality :) 

I tried using _onTreeExpandChange_ method, but I couldn't influence this anyhow.

**Expected behavior**
I would like rows to stay expanded after the action is invoked.

mbrn:
Hi @JKrymarys,

Can you share your code please. If you send data directly to component it would be. You should use data from your state.

JasonLunsford:
I'm also experiencing this issue. Our tree can have 3 or greater levels of depth and after the third level is reached - boom - branch collapses.

JKrymarys:
Hi Guys, 
So in my case code looks like: 

actions property in table:
` actions={[
                {
                  icon: "more_vert",
                  tooltip: "More",
                  onClick: this.handleActionsClick
                }
              ]}
`

while the function handleActionsClick looks like:
`handleActionsClick = (event, rowData) => {
    this.setState({
      anchorEl: event.currentTarget,
      rowData: rowData
    });
  };
`

Apparently "setState" causes collapsing of the tree. I tried also to ivoke setState in the "onClick" handler - the result was the same.

natalia-migas:
Hi, I have this problem as well and I find it really annoying. @mbrn you can check it in docs: https://material-table.com/#/docs/features/tree-data

When you check/uncheck a checkbox on the expanded child row, the whole table is rerendering and all rows are collapsing.

JKrymarys:
Table is apparently not remembering which rows were expanded.

I think that component should check it and pass to the re-render function.

natalia-migas:
Hi, refering to my last comment this issue should be labeled "bug" :) Are you gonna look at this problem, pls?

JKrymarys:
Hi @mbrn !
Any updates on this issue? :)

TheGoodSpinach:
Hey, running into this issue on one of our tree grids, specifically when filtering. You can replicate it here:

 

https://codesandbox.io/s/keen-bassi-nuecq?fontsize=14

 

Steps to reproduce:

1. Type "foo" into filter

2. Expand "foo" row

3. Select "bar" row checkbox

4. "foo" row collapses again

 

[...]

PR #1377, opened 2019-11-23
fix issue #743

res63661:
## Related Issue
#743 

## Description
Trying to prevent erroneous collapse of expanded nodes when events fire from within the row

## Related PRs

branch | PR
------ | ------
other_pr_master | [link]()

*

## Additional Notes
I can see that the goal was to re-apply expansions later in the function.  If my fix is too sweeping then, if you can give me some backstory and clear goals, I may be able to apply them for you and re-submit.


mbrn:
Hi @res63661 

Thanks for your effort. But this would affect on search/filter changes. We want to expand groups that hits the search/filter term. 

Instead of removing code you must find a difference between search/filter changes. 

I am closing PR for now. You can reopen it if you find a solution.

PR #1381, opened 2019-11-24
fix initialPage for remote data

damianszocik:
## Related Issue
#964

## Description
Making initialPage options property to work with a remote data query. 

Added `remoteDataMode` property and `setRemoteDataMode` method in data-manager to ommit pagination data split when a remote data query is set. 

material-table component is adjusted to store correct query object. initialPage is now included in query object when it's provided in options prop.

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -35,7 +35,7 @@ export default class MaterialTable extends React.Component {
           })),
         orderBy: renderState.columns.find(a => a.tableData.id === renderState.orderBy),
         orderDirection: renderState.orderDirection,
-        page: 0,
+        page: calculatedProps.options.initialPage || 0,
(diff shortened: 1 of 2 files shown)


pedro-lb:
Since this hasn't been merged in a while we've created a fork to fix this issue:

https://www.npmjs.com/package/material-table-hotfix-initial-page-remote-data

mbrn:
Hi @damianszocik 

Thanks for your effort. But i don't want to move remoteData flag to inside DataManager. DataManager should not be aware of data source. 

Can you consider this rule and find a new solution, please?

Also can you solve conflicts?

mbrn:
No action for more than 10 days. I am closing the pr. You can open it after an update.

Issue #1326, opened 2019-11-12
Actions `disabled` prop is not respect in 1.54.0 and 1.54.1


callmeberzerker:
**Describe the bug**
The `disabled` prop is not respected on the `Action` object.

**To Reproduce**
1. Go to any of the examples for actions
2. Edit the action to have `disabled: true` prop
3. See that the action is not disabled

**Expected behavior**
`disabled` prop is respected.

**Screenshots**
n/a

**Desktop (please complete the following information):**
 n/a

mvitkin:
I am not able to reproduce this bug, neither in my local environment, nor in the examples for actions. Adding `disabled: true`, does seem to disable the buttons (I am no longer able to click on them).

Can you share an example where you see this occurring?

IraSkyx:
I can reproduce this.

CodeSandbox : https://codesandbox.io/s/material-table-issue-with-disabled-4iioi

It seems to occur only when passing a function as action 
`rowData => ({
            icon: React.forwardRef((props, ref) => (
              <Check {...props} ref={ref} />
            )),
            disabled: true
          })`

Doesn't work in 1.54.1 

While : 
`{
            icon: React.forwardRef((props, ref) => (
              <Check {...props} ref={ref} />
            )),
            disabled: true
          }`

Does work even in 1.54.1

msanmartin-cloudhesive:
Having the same issue with 1.54.1. I'm using the sample version from the documentation on Conditional Actions, which also doesn't seem to be working [Conditional Actions Example](https://material-table.com/#/docs/features/actions)

[CodeSandbox Example](https://codesandbox.io/s/optimistic-leavitt-vh3hc?fontsize=14)

Behavior is the same that @IraSkyx describes. It doesn't work when passing a function and when passing 'disable' prop in an object it works but affects every row

Shurpop:
Hello!

After update from version 1.50.0 to 1.54.1 it doesn't work.

Eplumecocq:
I also have the same issue in 1.54.1

Tainan404:
I have the same bug when I use a function to actions.

https://codesandbox.io/s/react-example-zumc6?fontsize=14&hidenavigation=1&theme=dark

dmytro-mavrosaus:
Could anybody merge this PR?

yoniabrahamy:
I join @dmytro-mavrosaus request

hikmert:
Hi,
[...]

PR #1383, opened 2019-11-25
Change disable declaration to get disable condition from fuction too.

Tainan404:
Bug reported in issue #1326. 

What happens is that the disabled variable is declared before the function is executed if the action is a function, thus not assigning disable from the function.

### DIFF
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -9,7 +9,6 @@ import Tooltip from '@material-ui/core/Tooltip';
 class MTableAction extends React.Component {
   render() {
     let action = this.props.action;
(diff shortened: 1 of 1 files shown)


gabolucuy:
please any updates on this pr?

welljsjs:
@mbrn

gabolucuy:
@demkonst if you really need the solution in the mean while you can override de  _src/components/m-table-action.js_ with the solution on this pr

callmeberzerker:
Guys, we really need this as well. :) 

The whole disabled feature is kinda borked in the current version...

mbrn:
Hi

Thanks. I already merged similar pr. Thanks for your effort.

Issue #1321, opened 2019-11-12
isEditable conditional is broken in version 1.54.1


bilix:
**Describe the bug**
isEditable is not working in version 1.54.1. It works fine in version 1.54.0. If the condition is set to false, you can edit the row anyway.

**To Reproduce**
```
editable={{               
    isEditable: rowData => false,
    onRowUpdate: (newData, oldData) =>
        new Promise((resolve, reject) => {
            setTimeout(() => {
                {
                      /* const data = this.state.data;
                      const index = data.indexOf(oldData);
                      data[index] = newData;                
                      this.setState({ data }, () => resolve()); */
                 }
                 resolve();
                 }, 1000);
              }),
         }}
```

**Expected behavior**
The edit action button is disabled

ogenodisho:
(This repo needs some unit tests)

georgehazan:
I confirm. not only isEditable, but also isDeletable too

thuryn:
I can confirm it also

nathann282:
this issue is still in version 1.54.2

jfocco-IUGO:
I can confirm it also, isEditable is not working yet...

thelazydogsback:
Also running into this issue

evblurbs:
Also experiencing this. I used `patch-package` module to apply the change for my project.  Thanks @benrawn for the fix.

william13murphy:
I am still experiencing this issue in 1.57.2. Can anybody else confirm?

LuisPinaSoares:
@william13murphy i do not confirm.
Just updated my version to 1.57.2 and its seems to be working

`          editable={{
            isEditable: rowData => selectedSubst.includes(rowData.id),
            onRowUpdate: (newData, oldData) => {},
          }}`

william13murphy:
@LuisPinaSoares Thank you for your response. isEditable is working as intended!!

Thank you package managers/contributors for a wonderful library.

PR #1400, opened 2019-11-30
Fix issue #1321: broken isEditable and isDeletable options

benrawn:
## Related Issue
isEditable conditional is broken in version 1.54.1 #1321

## Description
Fixes a bug that is causing the `editable` property's `isEditable` and `isDeletable` options to be ignored.


Androsv96:
A new version will be published with this solution?

Issue #1402, opened 2019-12-03
Column lookup attribute and type attribute do not play well together


macyabbey:
**Describe the bug**
A column configured with a "lookup" prop results in a string typed value in the submit data, regardless of the "type" attribute of the column.

**To Reproduce**
Configure a column with type "numeric". Give the column a type of "numeric".
Configure the table to be editable.
Inspect the data passed to the onRowAdd function.
Notice the type of the column is not numeric, it's a string.

**Expected behavior**
The type of the field's value should match the type of the column.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: OS X
 - Browser: Chrome
 - Version Version 78.0.3904.108 

**Additional context**
This should be extended to all possible types.

Example code demonstrating the issue.

```

    let dogTypeLookup = {
        0: 'Golden Retriever',
        1: 'Labrador'
    };

    return (
        <MaterialTable
          options={{
            pageSize: 15
          }}
          columns={[
            { title: 'Dog type', field: 'dog_id', lookup: dogTypeLookup, type: 'numeric', readonly: true}
          ]}
          title=""
          editable={{
            onRowAdd: (data) => {
                //When selecting Golden Retriever

                //Actual Result
                //data.dog_id === 0 is false
                //data.dog_id === "0" is true

                //Expected Result
                //data.dog_id === 0 is true
                //data.dog_id === "0" is false
                return Promise.resolve(data.dog_id === parseInt(data.dog_id));
            }
          }}
        />
    );

```

miktap:
Valid issue for other than lookup columns as well, see #1450

PR #1403, opened 2019-12-03
After retrieving the value out of the Select component when editing a…

macyabbey:
… field with the lookup property, attempt to coerce it to the right typeif type is also set on the column.

## Related Issue
#1402 

## Description
Enforce types of data values based on the column's type when a column uses the lookup feature.

## Related PRs
None

* Columns that use the lookup and type attributes in tandem.

## Additional Notes
I don't know enough about the project to know if this change will have unintended side effects. I tested it with my local project and it works for the issue I've identified.

### DIFF
--- a/src/components/m-table-edit-field.js
+++ b/src/components/m-table-edit-field.js
@@ -13,12 +13,36 @@ class MTableEditField extends React.Component {
     return props;
   }
 
+  coerceLookupType(type, value) {
+    if (type === 'numeric') {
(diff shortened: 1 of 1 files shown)


mbrn:
Hi @macyabbey 

Thanks for your effort. But this might be some side effects. Because some developers may use exact values to compare or anything else. 

So i don't want to merge this pr. 

Thanks for your understanding.

Issue #1048, opened 2019-08-29
Set focus to search field


yakuphanbilgic:
Is there any way that I can set the focus to the search bar on the table, so that as soon as the page renders users can type what they want to search?

oze4:
There is no option for this but you do have the ability to override components (in this case, the Toolbar component).. or you could even use your own custom search field outside of the table - there are many ways to solve this - all of which are not a simple `focus={true}` prop.

MaxInertia:
@oze4 We can override the search bar, but how do we get the table to recall our remote data function? I'm currently looking for a way to do this. Will be making an issue if I can't find anything in the issues about this.

harveyappleton:
@yakuphanbilgic  - I ended up wrapping the MUI table in a div with a unique id (used lodash to generate and store in a useRef hook to keep it "constant"). 

you can then use a simple querySelector to focus the input 

`        let input = document.querySelector(
          `.${id.current} input[type='text'].MuiInput-input`
        );
        if (input) input.focus();`

its a bit hacky, but it worked for me, and you can get it to focus after each re-render using useEffect hooks

oze4:
Another way you could do this is by overriding the toolbar, wrapping in div, giving that div an ID, and setting focus that way.

You also may be able to do this using refs? Not sure..

### [Check out a live demo here](https://stackblitz.com/edit/react-kpexmh?embed=1&file=index.js)

```javascript
// Main bits from demo
import React, { useState, useEffect, useRef } from "react";
import { render } from "react-dom";
import MaterialTable, { MTableToolbar } from "material-table";
import tableIcons from "./TableIcons.js";

const originalData = ["Rock", "Paper", "Scissors"].map(word => ({
  id: Math.floor(Math.random() * 300),
  name: word
}));

const columns = [
  { title: "Id", field: "id" },
  { title: "Name", field: "name" }
];

function App() {
  /**
   * SET THIS TO WHATEVER ID YOU NEED.. could even have this in state..etc..
   *
   * May be able to do this using a ref on the table or toolbar component(s)?
   * ...instead of querying the DOM outside of React, which isn't really best practices.
   * // const tableRef = useRef();
   * // const toolbarRef = useRef();
   */
  const TOOLBAR_ID = "some__unique__hacky__id";

  const [data, setData] = useState(originalData);

  useEffect(() => {
    const searchBar = document.querySelector(`#${TOOLBAR_ID} input`);
    if (!searchBar) return;
    searchBar.focus();
[...]

PR #1406, opened 2019-12-03
Added searchAutoFocus to options

pradej:
## Related Issue
https://github.com/mbrn/material-table/issues/1048

## Description
This PR adds a possibility to set autoFocus on search input.


xiaomaini:
I would personally add a new prop called  ’searchProps‘ to avoid creating lots of search related props. What do others think?

mbrn:
Hi @pradej 

Thanks for your effort. This wourl be very useful and simple feature. 

@xiaomaini , We can add a overriding option for SearchField to make search feature more customizable. But this minor feature is acceptable for search.

Issue #1274, opened 2019-10-31
Add hook when detailPanel is opened or closed


dragonfriend0013:
**Is your feature request related to a problem? Please describe.**
I am needing to update my height in a SwipableView component whenever a detail panel is opened or closed.  Is there a possibility to add a onDetailPanelOpen and onDetailPanelClose callback so that we can provide a function to do additional work when this event happens?

**Describe the solution you'd like**
When a detailPanel is opened, trigger onDetailPanelOpen optional user provided function
When a detailPanel is closed, trigger onDetailPanelClose optional user provided function

**Describe alternatives you've considered**
none, we get no event when the panel is opened or closed

**Additional context**
I am using my table within SwipableView component and need to tell SwipableView to update its height due to the content of its view had changed due to the detailPanel being opened or closed.

bmz1:
This feature is exactly what I need

bmz1:
Created a pull request on this issue:
[#1452 ](https://github.com/mbrn/material-table/pull/1452)

Example:

`detailPanel={[
          {
            tooltip: 'Show order details'
            render: rowData => {
            // render data
},
          },
          {
            icon: 'account_circle',
            tooltip: 'Show Account details',
            onClick: rowData => {
              console.log(rowData)
             // onClick action e.g. fetch data
            },
            render: rowData => {
              // render something
            }
          }
        ]}`

TJHdev:
I was looking for this exact feature this morning. Thank you very much for adding it in!

psykzz:
This is possible already in onRowClicked, you have the `togglePanel()`. You can trigger your own event from inside that.

```js
onRowClick={(event, rowData, togglePanel) => togglePanel()}
```

Not the most direct path, but could unblock you for now.

bmz1:
I tried that, but it didn't work for me. If somebody needs this feature quickly, the forked package can be installed via npm.

bbatalla92:
It would also be nice to have a way to check if a rows detail panel is open. Possibly a flag on the data row object.

PR #1452, opened 2019-12-14
add onclick event to detailPanel

bmz1:
## Related Issue
[`#1274`](https://github.com/mbrn/material-table/issues/1274)

## Description
Added onClick handler to detailPanel array. Great to make API calls when clicking on a detailPanel icon

* renderDetailPanelColumn
* detailPanel props


mbrn:
Hi @bmz1 

Thanks for your effort. But you can call your api on your detail panel components` componentDidMount function. 

In addition we can add new hooks for detail panel like onExpand, onCollapse. If you want you can develop these features and they would be useful. But onClick is not understandable at first. 

I am closing issue now. If you develop more as i want, you can reopen it. Thanks :)

Issue #1291, opened 2019-11-05
Ability to choose variants


ATLUSio:
**Describe the solution you'd like**
It would be ideal to be able to choose the variants for the table's filters and search bar.

**Describe alternatives you've considered**
Manually re-creating the css in a theme

**Additional context**
Something like the "contained" textfields in this picture (but would extend to all the fields):
![image](https://user-images.githubusercontent.com/11895359/68233256-d09ef380-ffcc-11e9-9789-4b86a9e33b47.png)

I would imagine it as a prop for the MaterialTable, e.g.:
```
<MaterialTable
    ...
    variant="outlined"
    ...
/>
```

Issue #1061, opened 2019-09-03
Lookup dropdown is positioned slightly to the bottom compared to normal filter


serendipity1004:
https://material-table.com/#/docs/features/filtering

If you look at the bottom most example, the lookup column has slightly less padding at the bottom which positions the dropdown lower than default filters. Need to fix this as it looks very ugly. Meanwhile, is there a quick fix for it?

BrunoYTanaka:
+1

kthyer:
@serendipity1004 This was bugging me as well but I found a quick fix until there is an official fix for this.

For each column that has the dropdown filter add

`filterCellStyle: {
              paddingTop: 1
            }`

This fixed it on chrome and firefox for me.

tobilarscheid:
thanks @kthyer - not perfect, but good enough as a workaround. unfortunately, this stops me from digging deeper into it and probably finding a proper fix ;-)

janderson215:
Thanks for pointing us in the right direction for a quick fix, @kthyer. In Chrome, that actually leaves it one pixel lower than the filter inputs for me, so I believe a slight correction is in order:
`filterCellStyle: { paddingTop: 0 }`

PR #1476, opened 2019-12-27
Add the ability to chose the variant of the search text field

AchilleasMich:
## Related Issue
#1291 (Partially)

## Description
Added the ability to chose the variant of the search field by providing **searchFieldVariant** under the **options** props.
The possible values are _standard_, _filled_ , _outlined_.
If not provided the default value is standard

## Example usage
`
<MaterialTable>
...
  options={{
    searchFieldVariant: 'outlined',
   }}
</MaterialTable>
`
search_field_variant

* MTableToolbar component
* Demo application which was updated to showcase the change

## Additional Notes
There was also an attempt to provide the ability to chose variant for the filter fields. The problem was with the lookup filter which did not render the correct variant but kept using the standard one as shown in the picture below. 
problem_lookup
This might be related to #1061. Due to this inability to change the variant of the lookup filter, no change was made to the other filters to ensure consistency

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -114,6 +114,12 @@ class App extends Component {
               options={{
                 grouping: true,
                 filtering: true,
+                searchFieldVariant: 'outlined',
+              }}
+              localization={{
+                toolbar: {
+                  searchPlaceholder: "Outlined Search Field",
+                }
               }}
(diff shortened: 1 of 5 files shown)


kdbaker:
@mbrn Is it possible to get this approved?

PR #1518, opened 2020-01-13
Fix/initial page for remote data

kili-technology:
## Related Issue
Relate the Github issue with this PR using `#964`

## Description
Making initialPage options property to work with a remote data query.

## Related PRs

branch | PR
------ | ------
other_pr_master | [link]()

*

## Additional Notes


pedro-lb:
Since this hasn't been merged in a while we've created a fork to fix this issue:

https://www.npmjs.com/package/material-table-hotfix-initial-page-remote-data

mbrn:
Hi @kili-technology 

This is same with #1381 

And reply would be same:

Thanks for your effort. But i don't want to move remoteData flag to inside DataManager. DataManager should not be aware of data source.

Can you consider this rule and find a new solution, please?

Also can you solve conflicts?

PR #1537, opened 2020-01-17
(chore): update demo to functional components and react hooks

AugustoCalaca:
## Description
Update the demo to function components and react hooks.

None. This is only refactor


mbrn:
Hi @AugustoCalaca 

Thanks for contribution. But i am not sure using hooks just only in demo. We might think about refactor project fully. 

I want to invite you material-table slack community if you want. Please mail to me.

AugustoCalaca:
great sounds like a good idea
share the link to the material-table slack community, please

PR #1541, opened 2020-01-17
Add option to pass props to action IconButton 🎉

### LINKED ISSUE #1540: ClassName is not passed to action icon btn
```jsx
const useStyles = makeStyles(() => ({
  disableBtnHover: {
    "&:hover": {
      backgroundColor: "transparent",
    },
  },
}))

function CompanyUsersTable() {
  const classes = useStyles()

  return (
    <MaterialTable
      icons={muiUtils.tableIcons}
      columns={[
        { title: "NAME", field: "name" },
        { title: "EMAIL", field: "email" },
        { title: "ROLE", field: "role" },
        { title: "LAST LOGIN", field: "lastLoginAt" },
      ]}
      data={[
        {
          name: "John Doe",
          email: "hq@acme.com",
          role: "Account Owner",
          lastLoginAt: "53 minutes ago",
        },
      ]}
      title="Demo Title"
      actions={[
        {
          icon: () => (
            <Button variant="contained" size="small" className={classes.disableBtnHover}>
              Edit user
            </Button>
          ),
          iconProps: {
            className: classes.disableBtnHover,
          },
          onClick: (evt, rowData) => {},
        },
      ]}
    />
  )
}
```

Is this a bug? I'm trying to add a `<Button/>` and remove the hover effect from the parent button that material-table inserts so I can make it looks good and not like this:

Screen Shot 2020-01-17 at 7 56 45 PM

Class attribute is not set:
Screen Shot 2020-01-17 at 8 02 47 PM

alfonmga:
## Related Issue
Fixes #1540

## Description
This change allows to pass `props` to the `<IconButton/>` so we can tweak it (e.g: pass a `className` prop so we can modify the styles without need to modify material theme as currently you need to do if you want to do this)

* MTableAction
* Types

## Additional Notes

LGTM ;-)

### DIFF
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -49,6 +49,7 @@ class MTableAction extends React.Component {
           color="inherit"
           disabled={disabled}
           onClick={(event) => handleOnClick(event)}
+          {...action.iconButtonProps}
         >
           {icon}
         </IconButton>
--- a/src/prop-types.js
+++ b/src/prop-types.js
@@ -13,6 +13,7 @@ export const propTypes = {
     position: PropTypes.oneOf(['auto', 'toolbar', 'toolbarOnSelect', 'row']),
     tooltip: PropTypes.string,
     onClick: PropTypes.func.isRequired,
+    iconButtonProps: PropTypes.object,
     iconProps: PropTypes.object,
     disabled: PropTypes.bool,
     hidden: PropTypes.bool,
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { IconButtonProps } from '@material-ui/core/IconButton';
 import { IconProps } from '@material-ui/core/Icon';
 import { string } from 'prop-types';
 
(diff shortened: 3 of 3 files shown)


alfonmga:
[Documentation](https://material-table.com/#/docs/features/actions) should be also updated to include this change.

| Field  | Type  | Default  | Description  |
|---|---|---|---|
|  iconButtonProps | object  |   | Props for icon button as IconButtonProps from https://material-ui.com/api/icon-button/#iconbutton-api |

alfonmga:
> could you send an example of this working?

Here's a CodeSandbox URL: https://codesandbox.io/s/muddy-leaf-fdrk0

It works, but I f*cked up and it doesn't solve what I initially wanted:  be able to use a `<Button variant="contained" />` as an action button. 

To achieve this last we should have an option to allow to pass a custom button element[ instead of using a predetermined button.](https://github.com/mbrn/material-table/blob/36fb268fd6ed00f786235719fc9d9261ff6c51e3/src/components/m-table-action.js#L46-L55)

PR #1557, opened 2020-01-21
allow overriding of self-defined classes

JapuDCret:
## Description
MaterialTable is customizable via MaterialUI's theming.

But if you for example want the MaterialTable header to use your primary theme color, you can only use the in-line CSS style

```jsx
<MaterialTable
    {/* . . . */}
    headerStyle: {
        backgroundColor: CustomTheme.palette.primary.main,
    },
/>
```
it would be more flexible to allow custom-class overriding

```jsx
<MaterialTable
    {/* . . . */}
    headerClassOverrides: {
        header: classes.tableHeader,
    },
/>
```

## Related PRs
none

* index.d.ts
* PropTypes of
  * MaterialTable
  * MTableHeader
  * Pagination
  * Scrollbar

The affected components are all components that use custom Material UI classes.

## Additional Notes
By default MaterialTables own classes will still be used. But users can choose to override them.

### DIFF
--- a/src/components/m-table-header.js
+++ b/src/components/m-table-header.js
@@ -207,6 +207,7 @@ MTableHeader.propTypes = {
   detailPanelColumnAlignment: PropTypes.string,
   hasSelection: PropTypes.bool,
   headerStyle: PropTypes.object,
+  headerClassOverrides: PropTypes.object,
   localization: PropTypes.object,
   selectedCount: PropTypes.number,
   sorting: PropTypes.bool,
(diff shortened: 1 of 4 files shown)


leonardlib:
This works?

JapuDCret:
I solved this by wrapping my MaterialTable in it's own [ThemeProvider](https://material-ui.com/customization/theming/#theme-provider). As this is a more flexible solution and a less intrusive one, I closed this PR.

PR #1558, opened 2020-01-21
Style and date format options

ckeboss:
# Description
Removes some hard coded styles and allows more customization ability from props

* Props


mbrn:
Hi @ckeboss 

Thanks for your effort. But i want to remove material-ui dependency in time. And apply a effective styling options to all components. So i don't want to merge this pr for now. 

Thanks for your effort again.

ckeboss:
Sounds good. I would agree a library independent styling solution would be good.

PR #1573, opened 2020-01-27
fix(types): change void to any for promises return on editable callbacks

hyphaene:
## Related Issue
I haven't opened an issue for this, but current types do not work with react apollo hooks generated with graphql-codegen

See 
errors_types

Because the functions available through the hook have a stronger Promise return Type 

## Description
This PR is meant to enhance compatibility of this library with apollo react and graphql-codegen

* none

This will only affect types and **is not** a breaking change since it's more permissive, but imo the original type was restrictive and worked well by directly returning a Promise object, which does not cover use cases with third party librairies 

## Additional Notes
For now the workaround for using react apollo and graphql-codegen is to edit the types files directly in the node_modules folder

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -11,9 +11,9 @@ export interface MaterialTableProps<RowData extends object> {
   editable?: {
     isEditable?: (rowData: RowData) => boolean;
     isDeletable?: (rowData: RowData) => boolean;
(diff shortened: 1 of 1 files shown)


hyphaene:
Happy to help !

Le jeu. 6 févr. 2020 à 09:04, Mehmet Baran <notifications@github.com> a
écrit :

PR #1591, opened 2020-01-30
Feature/action style div

fr3sh:
## Related Issue
Style div action `#`1587

## Description
New prop option for style action inner div . Now you can add style to make vertical aligment of actions. Ex: { display: 'grid'} or set width to stack icon {width: }


RegisParpex:
Really good idea.
We will be able to align the content with option : actionsCellInsideStyle: { align-items: center; justify-content: center }

mbrn:
Hi @fr3sh 

Why we don't remove div outside of actions. So we can override actions however we want?

Please try removing div instead of adding new prop.

mbrn:
No action for more than 10 days. I am closing the pr. You can open it after an update.

Issue #1603, opened 2020-02-03
Management of Selected Rows through Properties


PeterJFerrarotto:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to control the selected rows from outside of the table, such that I can clear it or add selected rows programmatically. Also, the onRowSelected property function is not firing. (I'll fix it all in the pull request).

**Describe the solution you'd like**
I want to be able to pass in an array of objects representing the currently selected data, and then modify that array from an external source (other than the table itself) and be able to change what is selected.

**Describe alternatives you've considered**
This is not currently a feature. I will be creating a pull request for it.

PR #1604, opened 2020-02-03
Management of Selections through Properties

PeterJFerrarotto:
## Related Issue
#1603 

## Description
Allows for row selection to be managed from outside of the table itself.
Also properly fires `onRowSelected`, and will include whether or not the row selection has been changed to a true state, to allow for more granular operations.

Added parameter `checked` to `onRowSelected`, and `rowChecked` to `onSelectionChanged`.

* Row selection
* `onRowSelected`
* `onSelectionChanged`

## Additional Notes
If a property for `selectedRows` is supplied, and neither the `onRowSelected` nor the `onSelectionChanged` property function alters the state value, the table will reflect the change, but the state won't change. If the property `externalSelection` is provided, it will not be able to change the selection whatsoever.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -37,6 +37,7 @@ class App extends Component {
   colRenderCount = 0;
 
   state = {
+    selectedRows: [],
     text: 'text',
     selecteds: 0,
     data: [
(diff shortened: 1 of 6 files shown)


mbrn:
Hi @PeterJFerrarotto 

Thanks for your effort. But this PR contains a lot of change on components and some of them are not suitable for our roadmap. In time, i want to decrease complexity of code(refactor). Then i work on issues like this. 

Thanks for your effort again.

Issue #1272, opened 2019-10-31
SearchText reset when parent component renders


re-anderson:
**Describe the bug**
The search box gets cleared if the parent component of a table gets rendered. For example, when a useState hook is triggered.

**To Reproduce**
1. Create a component with a useState hook and MaterialTable as a child
2. Type in a search term in the table's search box
3. Trigger the hook
4. See that search text gets cleared, along with resetting table results

**Expected behavior**
The search text is persisted on renders of parent component

**Screenshots**
![search-cleared](https://user-images.githubusercontent.com/57200445/67956462-22f1a600-fbc2-11e9-8227-1b562577bf3f.gif)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version 78.0.3904.70

**Additional context**
This seems to be related to pull request #1225. I tested by changing line 64 of material-table.js to `props.options.searchText && this.dataManager.changeSearchText(props.options.searchText);` and the search text was persisted.

edit: specify useState hook

re-anderson:
Closing since this should've been fixed with PR #1298

Issue #491, opened 2019-04-21
Filtering states reset upon rerendering remote table


jakeleventhal:
**Describe the bug**
When re-rendering a remote table, all the filter states reset - meaning the search text gets reset.

**To Reproduce**
1. https://codesandbox.io/embed/j2jo7j91ky
2. Enter text into one text field
3. The other filter goes away, but the text disappears

**Expected behavior**
The filter text should remain

**Additional context**
This is relevant because firestore has limitations as to how many conditions can be on a query - so you want to only allow the user to be able to filter certain fields

mbrn:
Hi @jakeleventhal 

This is not a bug. You used columns directly. I advise you to use them in parent components' state. So tableData of columns will not be disappear when you setState in data function.

dalenhughes:
Are you sure this isn't a bug? You can see here that my table has the state of my filters but doesn't display the 'nar' filter in its appropriate filter field. Disappears after the search happens.

<img width="1360" alt="Screen Shot 2019-05-13 at 2 26 32 PM" src="https://user-images.githubusercontent.com/11064822/57663011-fb246c80-75a6-11e9-8fb0-9ac18e48fbc8.png">

jakeleventhal:
This is the problem i was having @dalenhughes

mbrn:
@dalenhughes 

Can you share your code please? If possible with a codesandbox.

dalenhughes:
@mbrn When you say 'You used columns directly. I advise you to use them in parent components' state' what do you mean by 'you used columns directly'? Do you mean that we can't use a parent's components state to dynamically create the columns like this?  Here is my render function:
`
```javascript
render() {
        const { classes, title } = this.props;
        const { data, headers, meta, columnNames, columnLinks } = this.state;

        let columns = _.zipWith(headers, columnNames, (h, n) => { return { title: h, field: n } });
        _.forEach(columns, c => {
            if (_.includes(columnLinks, c.title)) {
                c.render = (value) => {
                    return (
                        <Link to={`/reports/${columnLinks[c.field].report}?per_page=20&page=1&filter{${columnLinks[c.field].column}}=${value}`}>{value}</Link>
                    );
                };
            }
        });

        return (
            <div>
                <Paper className={classes.root}>
                    <MaterialTable
                        title={title}
                        options={{
                            filtering: true,
                            columnsButton: true,
                        }}
                        data={this.handleTableChange}
                        columns={columns}
                    />
                </Paper>
            </div>
        );
[...]

PR #1611, opened 2020-02-04
fix: remove tableData property from column in check phase

giacomocerquone:
## Related Issue
#491 
but actually many more like #1272 (partially solved) #1265

## Description
This PR introduces a fair equality check between columns when using remote data.
Without this pr, while checking in component did update if the columns changed, `this.props.columns` and `prevProps.columns` assumed different forms hence the deep equal returned false for the first time causing a rerendering of the entire table losing every setting set, and true at the second time.

This happens because through the data-manager the tableData property is added only in a second time to this.props.columns.

* the entire material-table.js component is affected


giacomocerquone:
I've also removed the equality check on the data only when isRemoteData is true, since checking for equality between two functions make no sense

giacomocerquone:
thanks @mbrn

Issue #555, opened 2019-05-09
Uncaught TypeError: Cannot read property 'tableData' of undefined after first page


exp0nge:
**Describe the bug**
From page 2 and onwards, when you click on the detail pane / chevron, the table errors out with: `Uncaught TypeError: Cannot read property 'tableData' of undefined`

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - Version [e.g. 22] 1.32.0

**Smartphone (please complete the following information):**

**Additional context**

exp0nge:
@mbrn for reference the code: https://github.com/mbrn/material-table/pull/175#issuecomment-459372239

Domino987:
Close in favor of #175

PR #1629, opened 2020-02-09
fix search bug

gyn7561:
## Related Issue
https://github.com/mbrn/material-table/issues/555

## Description
line 723  if [rowData.tableData.path] has same value , line 732 [pointer.tableData.markedForTreeRemove = true] will throw exception.


mbrn:
Hi @gyn7561 

I could not reproduce the bug. Can you create a sandbox or explaing all steps to reproduce the bug. 

Thanks

mbrn:
No action for 9 days. I am closing the pr. You can open it after an update.

PR #1638, opened 2020-02-11
Resolving the issues  #1644 and also, updating Material-UI to the lastest version v4.9

orestes22:
## Related Issue
#1644 Resolving issue with PageSize that change to original value if a props in option dynamically
Update the material ui to the version 4.9. 

## Description
Simple words to describe the overall goals of the pull request's commits.

## Related PRs

branch | PR
------ | ------
other_pr_master | [link]()

*

## Additional Notes


src/material-table.js
```diff
@@ -83,7 +83,11 @@ export default class MaterialTable extends React.Component {
     isInit && this.dataManager.changeOrder(defaultSortColumnIndex, defaultSortDirection);
     isInit && this.dataManager.changeSearchText(props.options.searchText || '');
     isInit && this.dataManager.changeCurrentPage(props.options.initialPage ? props.options.initialPage : 0);
-    (isInit || this.isRemoteData()) && this.dataManager.changePageSize(props.options.pageSize);
+    
+    if(isInit && this.isRemoteData()) {
```
mbrn, line 87: I am not sure this is true condition. I know that we have to find a better solution instead of a workaround.

package.json
```diff
@@ -49,7 +49,7 @@
     "@babel/plugin-transform-runtime": "7.1.0",
     "@babel/preset-env": "^7.2.0",
     "@babel/preset-react": "7.0.0",
-    "@material-ui/core": "^4.0.1",
+    "@material-ui/core": "^4.9.2",
```
mbrn, line 52: We should test before merging this.

mbrn:
Hi @orestes22 

Thanks for contribution. But i can't merge this for reasons that i mention on code. Thanks for understanding.

package.json
orestes22: Right now, Im working with that version for several month and working fine. I have a fork where Im doing my own changes.

src/material-table.js
orestes22: it is your call, and you could find a better solution, but right now it is a big issue for people are using the table in remote mode.

PR #1643, opened 2020-02-12
Allow `React.ReactNode` for certain localization strings

Svish:
## Description
Allow passing in React components as localization options. This allows for example changing typography, "tagging" strings with data attributes for testing, etc.


Domino987:
Maybe set the type to `string` and `React.Node` so that user know that the default should be a string and only if something like changing for testing that needs to be used.

Svish:
`string` is already part of `ReactNode`. And there's not really any reason why it should be `string` "by default".

Domino987:
Yes I know. But you are right. So looks good.

Issue #1121, opened 2019-09-21
Details dropdown spaced out too much with actions


jakeleventhal:
**Describe the bug**
If you have actions and a details panel, the dropdown for the details panel gets spaced out way too far. The details dropdown icon should be essentially right next to the actions buttons
<img width="544" alt="Screen Shot 2019-09-21 at 1 36 07 PM" src="https://user-images.githubusercontent.com/15133828/65378950-21ed6080-dc75-11e9-9b78-8bddcba4b2e3.png">

**To Reproduce**
Create a table with actions and a details panel.

**Expected behavior**
Icon should be next to the actions without giant space

**Screenshots**

ingvaldlorentzen:
I was experiencing a similar issue, but only when columns are hidden. I was not able to reproduce the issue without hidden columns, did you use hidden columns?

Anyways, I've fixed my issue in #1666

jakeleventhal:
not stale

ingvaldlorentzen:
> not stale

My fix has been merged. So this issue should be closed, no?

PR #1666, opened 2020-02-19
Fix: Action/Detail/Tree spacing with hidden columns

ingvaldlorentzen:
## Related Issue
#1121 

## Description
Fixes a bug that would cause the iconbutton (action/detail/tree etc) on the right side to take on the "extra" space "leftover" when hiding columns. See the related issue for how this looks.   
This is fixed by filtering out hidden columns when checking the length for the CSS `calc`. 

## Related PRs
None that I could find.

* data-manager.js

### DIFF
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -19,7 +19,7 @@ export default class DataManager {
   treeDataMaxLevel = 0;
   groupedDataLength = 0;
   defaultExpanded = false;
-  
+
   data = [];
   columns = [];
 
@@ -57,7 +57,7 @@ export default class DataManager {
     this.filtered = false;
   }
 
-  setColumns(columns) {    
+  setColumns(columns) {
     const undefinedWidthColumns = columns.filter(c => c.width === undefined);
     let usedWidth = ["0px"];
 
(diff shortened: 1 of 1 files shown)


gaurav-agarwal27:
@mbrn how long it can take to merge and close this PR ? Really need this

Shmew:
I ran into this problem earlier this week. You can get around this in the mean time by doing a theme override of the `.MuiTableCell-paddingNone` and `.MuiTableCell-paddingCheckbox` by setting the width to `calc((100% - 0px) / 16)`

raethlein:
I stumbled on the same issue today. Great to see that a PR is already open! 

I would slightly modify the code and apply the hidden test already in line 61: 
` const undefinedWidthColumns = columns.filter(c => c.width === undefined && !c.hidden)`
as there filtering is already applied.

simondriesen:
> I ran into this problem earlier this week. You can get around this in the mean time by doing a theme override of the `.MuiTableCell-paddingNone` and `.MuiTableCell-paddingCheckbox` by setting the width to `calc((100% - 0px) / 16)`

If you are using the detailPanel prop this solution will break the detailPanel button width. This theme override fixed it for me:

```
MuiTableCell: {
  alignLeft: {
    width: 'auto !important',
  },
},
```

ingvaldlorentzen:
> Could you remove the changes to white space, so that the changes that this PR is actually doing are more visible?
Fixed.
> Also since the filter is already applied in line 61, could you move the filter up to that and see if that works too?
Good catch! I moved it to the existing `filter()`, works well in all my environments!

PR #1671, opened 2020-02-20
Fix: dateTimePickerLocalization prop is not working

siebenrene:
## Related Issue
Relate the Github issue with this PR using #1129 

## Description
dateTimePickerLocalization prop is not working because a prop 
was not set correctly m-table-edit-row. In addition while adding new rows the prop was 
not propagated to the datePicker component. Edit rows worked fine.

Solution: Fixed incorrect variable in m-table-edit-row. Fixed missing
propagation of dateTimePickerLocalization in m-table-body.

### DIFF
--- a/src/components/m-table-body.js
+++ b/src/components/m-table-body.js
@@ -159,7 +159,7 @@ class MTableBody extends React.Component {
             icons={this.props.icons}
             key="key-add-row"
             mode="add"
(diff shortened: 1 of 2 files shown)


siebenrene:
Conflicts with #1664 - after review fix is already included in pull #1664

PR #1683, opened 2020-02-21
Fixes Cells aligned

### LINKED ISSUE #1678: Bug in show and hide columns if the property selection is true
**Describe the bug**
There is a bug with the table that the columns align to right when you hide a columns throw the menu in the toolbar. 

**To Reproduce**
1. Add to "option" the properties "selection:true" and "columnsButton: true"
2. After that click the menu in the toolbar and hide some columns and you will see how all the columns aligned to right.

Check this example https://codesandbox.io/s/issues-material-table-with-the-columns-3v70d

**Expected behavior**
The behavior have to be like when the selection property is false

**Screenshots**
https://codesandbox.io/s/issues-material-table-with-the-columns-3v70d

 - OS: [window]
 - Browser [crome]
 - Version [1.57.2]

**Smartphone (please complete the following information):**

**Additional context**

Domino987:
This fixes #1678 for if selection and columnsButton are enabled and columns are removed that the cells align to the right of the table.

## Related Issue
 #1678

## Description
This makes the layout fixed so that the cells are correctly aligned for removed columns.
The selection column does not span unnecessary. 

* Material-Table

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -538,7 +538,7 @@ export default class MaterialTable extends React.Component {
   }
 
   renderTable = (props) => (
-    <Table style={{ tableLayout: (props.options.fixedColumns && (props.options.fixedColumns.left || props.options.fixedColumns.right)) ? 'fixed' : props.options.tableLayout }}>
+    <Table style={{ tableLayout: (props.options.fixedColumns && (props.options.fixedColumns.left || props.options.fixedColumns.right)) || (props.options.selection && props.options.columnsButton) ? 'fixed' : props.options.tableLayout }}>
       {props.options.header &&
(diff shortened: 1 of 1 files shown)


orestes22:
@Domino987 This solution it is not a good idea because you will lose a responsive table using a mobile.

orestes22:
@Domino987 Hi, could you explain me why are you closing this if i commend you that your idea isnt a good solution when you are in mobile screen. 

The original issue persist

Best regards

Domino987:
Because since this solution does not work as I thought it might, I closed it. It should be moved into a issue so that maybe has an idea to fix it

Issue #980, opened 2019-08-13
Selecting row hides toolbar action menu


Dwight-D:
**Describe the bug**
When rows are selected, the toolbar.nRowsSelected message blocks the toolbar action menu. Even when options.showTextRowsSelected is false the toolbar menu still disappears, even though no message blocks it.

**To Reproduce**
1. Enable selection and disable selection text in options prop:
```
const options = {
      selection: true,
      showTextRowsSelected: false,
...
```

2. Notice that toolbar action menu is shown
3. Select one or more rows
4. Notice how toolbar menu disappears 

**Expected behavior**
Toolbar should stay regardless of row selection.
Possibly add additional option to force toolbar to stay on top, even if selection message is present (show both message and toolbar if there is space in the header).

**Screenshots**
No selection:
<img width="1626" alt="image" src="https://user-images.githubusercontent.com/25592807/62941987-9cee6c00-bdd7-11e9-9fd5-d76afb4eb3c2.png">

With selection and no selection message:
<img width="1627" alt="image" src="https://user-images.githubusercontent.com/25592807/62942012-aaa3f180-bdd7-11e9-8e8f-3ab6fc8c1e0e.png">

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser: Firefox Developer Edition, version 69.0b11 (64-bit)

bableves:
Any update on this? It's a pretty annoying bug

pixelass:
@mbrn can you please take a look at this? 

* Is this a bug or does it work as designed?
* Any suggestion how to make actions work with selections would be helpful

mbrn:
It seems like a bug. I will take a look it. But I prioritize the issues according to their reaction count. Please take a look at readme

pixelass:
Thank you for the fast reply. 
I need this feature but can use something custom until it's fixed.
I might look into it and try to open a PR if I can get it to work.

pixelass:
I'm not sure if this helps the other reporters. In my case I was able to find some undocumented  feature (I might also be blind) to support my desired behavior. 

```
actions={[{
    icon: 'edit',
    tooltip: 'Edit',
    position: "toolbarOnSelect",
    onClick: (event) => {
        console.log('Clicked edit action')
    }
}]}
```

This will show the action in the toolbar only if one or more rows are selected which is exactly what I was trying to achieve.

[...]

PR #1704, opened 2020-02-27
Updated actions to show toolbar actions even when selection is made

rajkthapa:
## Related Issue
`#980`
https://github.com/mbrn/material-table/issues/980

## Description
Actions icons which are set as toolbar position are not showing up when selection is made. Updated the code so that toolbar position actions are shown no matter what.

## Related PRs

branch | PR
------ | ------
other_pr_master | [link]()

*

## Additional Notes


mbrn:
Hi @rajkthapa 

Thanks for contribution. But this is not a bug. We don't want to show toolbar buttons when any rows selected. toolbar actions would be show when no rows selected.

Issue #1596, opened 2020-01-31
On row canceled editing?


pabichw:
Is it possible to catch when row editing was cancelled? So in other words, that "X" button had been pressed?

Issue #1313, opened 2019-11-10
How can I make the edit-mode field type of one column dependent upon the value of another column, without affecting other rows?


davidjmstewart:
**Summary**

I need the input field that appears when the user is in edit mode to be different depending on the value of a different column for that row. E.g. if the user is in edit mode and `Column X` has value `A`, then `Column Y` should be a text entry field. If, however, `Column X` has value `B`, then `Column Y` should be a dropdown select list. Importantly, this must not affect any other rows (i.e. the fact that `Column Y` might be a text entry field for this row should not affect the input type on any other row). 

Note: This is a repeat of my question [posted on StackOverflow](https://stackoverflow.com/questions/58776687/how-can-i-make-the-edit-mode-field-type-of-one-material-table-column-dependent-u)

**Mockup**

In case it wasn’t clear what I was asking for, here is a mockup that shows the functionality I’m after. 

***1. Target type is 'Group', so 'Target' field is numeric entry***
[![enter image description here][1]][1]

***2. Target type is 'Address', so 'Target' field is a select list of strings.*** Notice that the other rows entries are
unaffected (i.e. they are still numeric fields)
[![enter image description here][2]][2]

**More detail**

I am creating a lighting control app, and am using material-table as the means through which users specify what actions a button can take (it could be a single action, in which case the table would be a single row, or many actions, in which case the table would be many rows). They can either control a single device, which is stored in a database and should be selectable from a dropdown list, or a group of devices, in which case they manually enter the number of the group they wish to control.

**Code**

This code forms the basis of the mockups (above):

```
import React from "react";
import MaterialTable from 'material-table';

const devices = ["Device A (Stairs)", "Device B (Hallway)"];

export default class ControlSpecificationTable extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      columns: [
        { title: 'Step', field: 'step' },
        { title: 'Target Type', field: 'target_type', lookup: { 0: 'Address', 1: 'Group' } },
        { title: 'Target', field: 'target', initialEditValue: '0', lookup: { 0: devices[0], 1: devices[1] }},
        // { title: 'Target', field: 'target', initialEditValue: '0', type:'numeric'},

        {
          title: 'Control Type',
          field: 'control_type',
          lookup: { 0: 'Button', 1: 'Slider', 2: 'Arrows' },
        },
        {
            title: 'Command type',
            field: 'command_type',
[...]

Issue #1230, opened 2019-10-19
Callback after Cancel Button


jitendra9nov:
First of all I am new to react, so may be I am unable to figure out this by looking into document. 

I am trying to do some action after user selects '**Cancel (X)**' button during Add/Edit/Delete. I have added below code but this is getting called only after user selects **Save**

> editable={{
>         onRowAdd: newData =>
>           new Promise(resolve => {
>             setTimeout(() => {
>               resolve();
>               const data = [...state.data];
>               data.push(newData);
>               setState({ ...state, data });
>             }, 600);
>           }),
>         onRowUpdate: (newData, oldData) =>
>           new Promise(resolve => {
>             setTimeout(() => {
>               resolve();
>               const data = [...state.data];
>               data[data.indexOf(oldData)] = newData;
>               setState({ ...state, data });
>             }, 600);
>           }),
>         onRowDelete: oldData =>
>           new Promise(resolve => {
>             setTimeout(() => {
>               resolve();
>               const data = [...state.data];
>               data.splice(data.indexOf(oldData), 1);
>               setState({ ...state, data });
>             }, 600);
>           }),
>       }}

syed-sep23:
I have the same issue, is there any solution

triusis92:
Same here

itsmeganesh-cse-iiit:
Same issue, need on cancel callback and on add (+) free action callback, please let us know in case you find anything ..

Dobata7:
up

gjergjk71:
up

NikhilChugh:
You can do the workaround like in the code below

` <MaterialTable
        title=""
        data={tableData}
        style={{
          marginBottom: "8px",
        }}
        columns={column}
        headerStyle={tableOptions}
        options={tableOptions}
        editable={{
          onRowUpdate: (newData, oldData) =>
            new Promise((resolve, reject) => {
              setTimeout(() => {
                handleUpdateAction()
                resolve();
              }, 1000);
            }),
        }}
        components={{
          EditRow: (props) => {
            return (
              <MTableEditRow
                {...props}
                onEditingCanceled={(mode, rowData) => {
                  alert('Do action on row update cancel click')
                }}
              />
            );
          }}
        }
        localization={{
          header: {
            actions: "",
          },
        }}
        detailPanel={(rowData, idx) => {
          return (
            <FileDetailPanel files={item.files} documentId={rowData.id} />
          );
        }}
      />`

Issue #674, opened 2019-06-05
Add onRowCancel function on editable tables


jakeleventhal:
**Is your feature request related to a problem? Please describe.**
I have a custom edit component that is a select dropdown with two options (A, and B)

My current idea is to have a state variable that is `currentRowOption` set to `''` initially. In my editComponent, before returning the select component, I set the state variable to the value of the row, if and only if the state variable is equal to `''`. I am able to to set the state variable back to `''` in the onRowUpdate function, but I need to be able to reset it if the editing is cancelled.

**Describe the solution you'd like**
A function for `onRowCancel`

jakeleventhal:
I found a way to do this using material tables props that gets passed into the onedit function.

I still think this would be useful

ulrezaj:
@jakeleventhal Can you post an example of how you got it to work please?

jakeleventhal:
What i did was not use a state variable and instead use the supplied props on change function like this
```javascript
<TextField
    onChange={(e) => props.onChange(e.target.value)}
    value={props.value}
 />
```

This works, but for something more detailed, the onRowCancel would be really helpful.

chitgoks:
This is actually not needed. you can override editable.EditRow property that returns MTableEditRow and override onEditingApproved and onEditingCanceled.
this applies to both add, update and delete.

dashName:
> onEditingCanceled

seems like onEditingCanceled is not recognized.Do you have an example?

chitgoks:
```
components={{
        EditRow: (props) => {
          return (
            <MTableEditRow
              {...props}
              onEditingApproved={(mode, newData, oldData) => {
              }}
              onEditingCanceled={(mode, rowData) => {
              }}
            />
          );
        }}
      }

```

dashName:
> ```
> components={{
>         EditRow: (props) => {
>           return (
>             <MTableEditRow
>               {...props}
>               onEditingApproved={(mode, newData, oldData) => {
>               }}
>               onEditingCanceled={(mode, rowData) => {
>               }}
>             />
>           );
>         }}
>       }
> ```

so here you are rendering a whole new component for Edit.I am using EditComponent for columns.I don’t want to render the whole table again

dashName:
> ```
> components={{
>         EditRow: (props) => {
>           return (
>             <MTableEditRow
>               {...props}
>               onEditingApproved={(mode, newData, oldData) => {
>               }}
>               onEditingCanceled={(mode, rowData) => {
>               }}
[...]

PR #1724, opened 2020-03-08
New Editable props: onRowAddCancelled and onRowUpdateCancelled has been added

HarshithMjh:
## Related Issue
#1596 
#1230 
#786 
#674

## Description
Sometimes, we want to make the edit-mode field type and values of one column dependent upon the value of another column, without affecting other rows. 
#1313 
During such situation, we provide value and onChange function from our component's state. Assume the user modified up-to third column and then cancelled the edit. Then we will have to reset the modifications done in our component's state. Hence, a function call is needed whenever user cancels the operation. Since cancel can be done in add and update mode, I have added two new props, onRowAddCancelled and onRowUpdateCancelled.

branch | PR
https://github.com/HarshithMjh/material-table/commit/1ecf3c0774427be93f70ed1320224ba6dc283048

There are no visual changes. Its just that the new props gets called, if the prop is supplied by the developer.


chitgoks:
up

jplabti:
I need that.

HarshithMjh:
Hi @jplabti ,
Can you please approve in review

HarshithMjh:
@jplabti Thank you.

Waiting for approving review by any reviewer with write access

gastongouron:
Looking forward to use this!

dashName:
is there any update on this?

ngelotte:
Any update on this? I could really use this functionality.

types/index.d.ts
```diff
@@ -14,6 +14,8 @@ export interface MaterialTableProps<RowData extends object> {
     onRowAdd?: (newData: RowData) => Promise<any>;
     onRowUpdate?: (newData: RowData, oldData?: RowData) => Promise<any>;
     onRowDelete?: (oldData: RowData) => Promise<any>;
+    onRowAddCancelled?: () => void;
```
mbrn: Actually, onRowAddCancelled & onRowUpdateCancelled has rowData Parameter

HarshithMjh: Thanks for approving and merging the code.
Shall I add the rowData parameter while calling onRowAddCancelled and onRowUpdateCancelled functions?

Mahagney: I try to use onRowAddCancelled but it does nothing.
Does it still work?

Issue #1707, opened 2020-03-01
Selection in grouping gives "NaN is an invalid value for the `marginLeft` css style property" warning


proohit:
**Describe the bug**
When I enable grouping of rows and selection, everytime I expand the group this warning shows up.

**To Reproduce**
I recreated a simple example from the grouping in the documentation and added selection, when the warning appears. [codesandbox](https://codesandbox.io/s/wonderful-keller-bkqhr?fontsize=14&hidenavigation=1&theme=dark) Try to expand a group and then check the console

**Expected behavior**
No warning when expanding group

**Screenshots**
![image](https://user-images.githubusercontent.com/46965017/75623438-3d588200-5baa-11ea-85c6-3db65b2f7ef1.png)

**Additional context**
To further troubleshoot,  I:
1. disabled grouping but enabled selection. No errors/warnings and expected behaviour
2. enabled grouping and disabled selection. Also no errors/warnings and expected behaviour

oze4:
Pretty sure either [this line](https://github.com/mbrn/material-table/blob/master/src/components/m-table-body-row.js#L57) or [this line](https://github.com/mbrn/material-table/blob/master/src/components/m-table-body-row.js#L60) is why this error is happening..

oze4:
@proohit I just tested a fix for this, confirmed that the fix works, and submitted a PR.

proohit:
@oze4 sounds good, thank you. Let's hope for @mbrn to accept the PR. Seems rather busy isn't he?

masbaehr:
Also having this issue, and the PR fix did work ;)

oze4:
***REMINDER FOR MYSELF***

Resubmit PR for m-table-body-row - these changes had to be removed from the PR referenced above.

Domino987:
Is this fixed?

oze4:
@Domino987 we can close it out if got these changes in a diff PR that hasn't been submitted yet.

kylenguyen1014:
@oze4 Is it fixed yet? I update material-table from 1.59.0 to 1.64.0 and the error is still there

PedrinhowVieira:
Yeah, I can confirm, the warning is still here.
"react": "^16.12.0"
"material-table": "^1.64.0"

masbaehr:
Unfortunately still experiencing! 

"material-table": "^1.68.0",
"react": "^16.13.1",


index.js:1406 Warning: `NaN` is an invalid value for the `marginLeft` css style property.
[...]

PR #1759, opened 2020-03-25
Update Issue Templates

oze4:
Resolve issue #1707 

## Related Issue
#1707 

## Description
When using selection with grouping, `this.props.level` appears to be undefined. `this.props.level` is used in calculations to determine some CSS properties, ultimately resulting in `NaN` and an error after expanding grouped column.

## Related PRs
NA

* m-table-body-row.js

## Additional Notes
NA


jroyleaz:
Any movement on this? It seems like a small but worthwhile bug fix.

oze4:
@jroyleaz I wouldn't count on this PR getting merged. This repo seems pretty much dead.

jroyleaz:
@oze4 Hmm, thanks for the heads up. I think I'll just fork it and do it myself, then. I'd like to expose more of the internals so you can change sorting, grouping, etc from actions. And maybe cross link search between MT instances with a shared context. 👍

oze4:
@jroyleaz a local fork is prob the best route to be honest. They're *extremely* slow moving in this repo and not very responsive. It's sad because this repo produces a solid product.. It's even more unfortunate since this repo is referenced in the official Material UI docs as a table "you should check out".

jroyleaz:
@oze4 yeah, for sure. It's a great product that has been hugely helpful in my current project. I also found it through the material UI docs. 👍

oze4:
@jroyleaz I started a maintained for [here](https://github.com/oze4/material-table-core). You can also [find more here](https://material-table-core.com)! This change has been made in that repo.

oze4:
@mbrn I updated the Issue Templates in this PR - can you review?

Issue #888, opened 2019-07-26
Centering action button


FroggDev:
Hi,

The property **actionsCellStyle** apply to the **action <td>**
If i want to center the custom action button in the TD, I  need to access to the properties of the extra **<div style="display: flex;">** container added from the code.

BTW, i don't know what is the utility of this div.

When the div **<div style="display: flex;">** is removed ( using the browser console for example) I can correctly align action button using 
`actionsCellStyle:{justifyContent:'center'}`

Thanks.

pak11273:
It's hard to read what you wrote.  You can center an action button like this:

````javascript
options={{actionsCellStyle: {display:"flex", justifyContent: "center"}}}
````

**Comment by FroggDev** [comment]
![2019-08-06 11_23_02-Window](https://user-images.githubusercontent.com/3765812/62528117-9e9abb80-b83c-11e9-8b79-375cfcd6b27a.png)

Thanks, but as you can see the bottom line isn't aligned with your code

**Comment by FroggDev** [comment]
finally i got it:

`actionsCellStyle:{justifyContent: "center"},`

**Comment by chitgoks** [comment]
that is actually what i did. but no, it does not work

```
actionsCellStyle: {
            justifyContent: 'center'
          },
```
Even adding display: 'flex' or textAlign: 'center'

Please see image. the css instead gets applied to the cell but it does not center it because the buttons have a <div style="display: flex"> container.
![cellstyle](https://user-images.githubusercontent.com/3955470/74503084-c45ef680-4f2a-11ea-89b4-57fdff33ff77.png)

**Comment by trianta2** [comment]
The solutions above don't work, this issue should be reopened

**Comment by FroggDev** [comment]
You can try something tricky like

> actionsCellStyle:{justifyContent: "center",maxWidth: 90,width: 90}

**Comment by lucacattide** [comment]
In waiting for the official fix, set it like this:
```
actionsCellStyle: {
  display: 'flex',
  justifyContent: 'center',
  padding: 16,
  width: '100%'
}
```

arbaz1006:
Thanks @lucacattide, works fine for me.

JeffEmery:
I had to add a bit more padding and push the bottom margin...

    actionsCellStyle: {
        display: 'flex',
        justifyContent: 'center',
        padding: '24px',
        width: '100%',
        marginBottom: '-1px',
    },

chitgoks:
[...]

PR #1817, opened 2020-04-11
Move actionsCellStyle to div container

JeffEmery:
The `actionCellStyle` css style object is currently applied to the `TableCell` and not the child `<div>` element. Because of this, the styles do not allow formatting of the element such as `justifyContent: 'center'`  This simple edit moves the styles to the div. Hopefully this is the intended behavior. See [Issue #888](https://github.com/mbrn/material-table/issues/888)

## Related Issue
See [Issue #888](https://github.com/mbrn/material-table/issues/888)

## Description
Enable `actionCellStyle` to affect the styling of the cell's `<div>` directly.

## Related PRs

branch | PR
------ | ------
other_pr_master | [link]()

* [MTableBodyRow](https://github.com/mbrn/material-table/blob/master/src/components/m-table-body-row.js)

## Additional Notes
Very new to PRs so I apologize in advance if I've screwed something up.

### DIFF
--- a/src/components/m-table-body-row.js
+++ b/src/components/m-table-body-row.js
@@ -37,8 +37,8 @@ export default class MTableBodyRow extends React.Component {
     const actions = CommonValues.rowActions(this.props);
     const width = actions.length * CommonValues.baseIconSize(this.props);
     return (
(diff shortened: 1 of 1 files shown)


mbrn:
Hi @JeffEmery,

Thanks for contribution. But if we move style to inner component, other developers' style might be affected wrongly. So i can't merge this. We should find another solution.

pouryapb:
@mbrn why "& div" does not work?!

PR #1836, opened 2020-04-16
Allow using a ref on the table

brummelte:


mbrn:
We already have ref for table. This indicates page rather than table.

PR #1846, opened 2020-04-21
feat(typescript): update column.type definition

johannpinson:
## Description
This PR is to update the TypeScript definition which is incorrect for the colum.type field.
I update it accord to the PropType : 
https://github.com/mbrn/material-table/blob/3031eab70105df8d88c17984ca5cea878a9eae3f/src/prop-types.js#L52


Domino987:
@johannpinson Seems like there is a a merge conflict. Can you resolve that?

johannpinson:
Hello @Domino987, I resolved the conflict, it must be good now 😉

PR #1889, opened 2020-05-06
Feature/expose export button

raph90:
## Related Issue
I needed to be able to use the export functionality outside of the toolbar, so I moved the handler for the csv into the data manager. For those looking to customise the export functionality (e.g not having it be tied to an icon) this will be useful.

## Description
To move the csv export functionality into the data manager, and then create a handler on the MaterialTable component. This means that the handler, and the functionality can be accessed through the Material Table ref. 

I've moved around various bits of code but the core addition is this to MaterialTable:

` onHandleExport = () => {
    if (this.props.options.exportCsv) {
      this.props.options.exportCsv(this.props.columns, this.props.data);
    } else {
      this.dataManager.handleCsvExport();
    }
  };`

I've also added a defaultExportCsv prop to the MTableToolbar, but the functionality is unchanged, it's just a refactor. 

I've also added setters and properties to the data manager so that it has information about the csv props.

The toolbar - but users won't notice. 

## Additional Notes
My auto formatter means that there's lots of red and green here, but the places to look at are:
- the data manager, in particular the new handleCsvMethod and the new properties.
- the toolbar component, to see how I've adapted it to use the data manager method.
- the Material Table component to see how the props to the toolbar have changed

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -1,32 +1,32 @@
-import { Grid, MuiThemeProvider, Button } from '@material-ui/core';
-import { createMuiTheme } from '@material-ui/core/styles';
-import React, { Component } from 'react';
-import ReactDOM from 'react-dom';
-import MaterialTable from '../src';
+import { Grid, MuiThemeProvider, Button } from "@material-ui/core";
+import { createMuiTheme } from "@material-ui/core/styles";
+import React, { Component } from "react";
+import ReactDOM from "react-dom";
+import MaterialTable from "../src";
 import Typography from "@material-ui/core/Typography";
 
-let direction = 'ltr';
+let direction = "ltr";
 // direction = 'rtl';
 const theme = createMuiTheme({
   direction: direction,
   palette: {
-    type: 'light'
-  }
+    type: "light",
+  },
 });
 
 const bigData = [];
 for (let i = 0; i < 1; i++) {
   const d = {
     id: i + 1,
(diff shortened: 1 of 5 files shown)


alievrenkut:
I suggest that a single method which can be called outside should be enough.
Also @raph90 You should only change minimal lines of code for your change. Your formatting changes makes maintainer's review much harder. Could you create a new merge request with minimal changes?

raph90:
> I suggest that a single method which can be called outside should be enough.
> Also @raph90 You should only change minimal lines of code for your change. Your formatting changes makes maintainer's review much harder. Could you create a new merge request with minimal changes?

Hi Alievrenkut - Most of the code is the same, and will work exactly the same (i.e none of the changes are breaking). I've just moved where the export method is defined so that you can access it via the table ref. I can see over the weekend though if I can simplify it. 

My fork of material-table has the changes applied if you want to have a look, R

mbrn:
Hi @raph90 

Changes have a lot whitespace changes. This makes code reading very hard for me. And you should resolve conflicts. Can you consider please?

Thanks for your effort.

raph90:
Hi Mehmet, I'll try to do this this week - thank you for looking at the changes, Raph

mbrn:
Please resolve conflicts

mbrn:
I will close the PR. You can reopen after resolving all conflicts.

PR #1918, opened 2020-05-15
Table icons updated

ahmetaysit:
icons are visible for nextjs users


mbrn:
Hi @ahmetaysit 

Thanks for your effort. But we can not upgrade material-ui references by this way. Can you remove upgrades, please?

Issue #958, opened 2019-08-08
Search fails with error: Cannot read property 'tableData' of undefined


benjaminvdb:
**Describe the bug**
Searching fails for the provided data (see below) and component configuration with the following error:

```
Uncaught TypeError: Cannot read property 'tableData' of undefined
    at markForTreeRemove (:3000/static/js/1.chunk.js:217284)
    at :3000/static/js/1.chunk.js:217308
    at Array.forEach (<anonymous>)
    at DataManager.treefyData (:3000/static/js/1.chunk.js:217298)
    at DataManager.getRenderState (:3000/static/js/1.chunk.js:216646)
    at MaterialTable.<anonymous> (:3000/static/js/1.chunk.js:215903)
    at later (:3000/static/js/1.chunk.js:201658)
```

The page doesn't crash, but the following is shown in the console and the search results are incorrect.

<img width="577" alt="Screenshot 2019-08-08 at 16 50 20" src="https://user-images.githubusercontent.com/8875786/62713275-bb7ded00-b9fc-11e9-94bb-fe97083fcb4d.png">

**To Reproduce**
Enter any search term in the upper right search box using the code and data provided below.

Required data: [tree.json](https://pastebin.com/ryZ6AreA)

```
import React, { forwardRef } from "react";
import data from "./tree.json";
import MaterialTable from "material-table";
import Search from "@material-ui/icons/Search";
import ViewColumn from "@material-ui/icons/ViewColumn";
import SaveAlt from "@material-ui/icons/SaveAlt";
import ChevronLeft from "@material-ui/icons/ChevronLeft";
import ChevronRight from "@material-ui/icons/ChevronRight";
import FirstPage from "@material-ui/icons/FirstPage";
import LastPage from "@material-ui/icons/LastPage";
import Clear from "@material-ui/icons/Clear";
import Check from "@material-ui/icons/Check";
import FilterList from "@material-ui/icons/FilterList";
import Remove from "@material-ui/icons/Remove";
import AddBox from "@material-ui/icons/AddBox";
import DeleteOutline from "@material-ui/icons/DeleteOutline";
import Edit from "@material-ui/icons/Edit";
import ArrowUpward from "@material-ui/icons/ArrowUpward";

const columns = [
  { title: "Diagnose", field: "value", sorting: true, defaultSort: "asc" }
  // { title: "id", field: "id" }
];

const icons = {
  Add: forwardRef((props, ref) => <AddBox {...props} ref={ref} />),
[...]

PR #1932, opened 2020-05-21
Ensure path data is not processed multiple times

hodg9:
The function `addRow` calls itself to add the `dataRows` parent `addRow(parent)`. This causes a reprocessing of the parent row, depending on the order of the data data source. If the parent row now has another sibling its path will be updated with the wrong index due to `parent.tableData.childRows.length - 1`. 

The `addRow(parent)` line is needed to ensure that the parent rows `tableData` is configured. However, once configured it shouldn't need to be reprocessed and can be skipped.

I believe adding check before setting the path value will solve an issue I've seen when searching tree based data. The error occures in markForTreeRemove while it tries to walk the hierarchy using path. see issue #958

## Related Issue
#958 

## Description
Fix search feature when tree based data is used.

## Related PRs

branch | PR
------ | ------
other_pr_master | [link]()

*

## Additional Notes

### DIFF
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -701,7 +701,7 @@ export default class DataManager {
 
         addRow(parent);
 
-        rowData.tableData.path = [...parent.tableData.path, parent.tableData.childRows.length - 1];
+        if(!rowData.tableData.path) rowData.tableData.path = [...parent.tableData.path, parent.tableData.childRows.length - 1];
         this.treeDataMaxLevel = Math.max(this.treeDataMaxLevel, rowData.tableData.path.length);
       }
       else {


oze4:
For those that are interested, I have been maintaining a fork of this repo - this issue has been resolved within that fork. [You can check out a live demo here](https://oze4.github.io/material-table-core/#/demo/004) (fix included)

[You can read more here](https://github.com/oze4/material-table-core/blob/master/.github/MoreInfo.md) about why I created, and am maintaining, a fork of this repo.

[You can view the GitHub repo here](https://github.com/oze4/material-table-core)

[@material-table/core](https://www.npmjs.com/package/@material-table/core) - if you can import it from 'material-table' you can import it from '@material-table/core'...

We are not making drastic changes to the codebase. We are keeping things as close to "as-is" as possible - our main goal is to resolve open issues/PR's in the material-table repo, but within the fork (it doen't appear this repo is actively maintained and I got sick of spending my time writing PR's for this repo just for them to go unnoticed)..

mbrn:
Hi @hodg9 

Thanks for effort. But this issue is already resolved.

Hellysonrp:
Hi,
That issue is closed but not resolved.
It still happens to me.

I tried using this solution to see if it would work, but it didn't.
https://github.com/mbrn/material-table/blob/95845eeb72303202dfc4a3cd9d5e2830c7be15e0/src/utils/data-manager.js#L729
pointer is undefined

shailahir:
This issue is still open. Please someone help on this.

PR #1970, opened 2020-05-31
Add function for exporting table as PDF in addition to the CSV option.

jardon:
## Related Issue
None exactly.  I needed this functionality for another project and thought that it may be useful as a general feature.

## Description
This branch incorporates a new export option into the existing export menu that supports PDFs.  The exportPDF option utilizes the existing properties structure so that you can exportAllData and also override the export PDF function.

## Related PRs
None.

* Header
* Toolbar
* Table?

## Additional Notes
I haven't contributed anything to this project before.  I searched for a contributing guide, but I could not find it.  I can fix whatever is needed to meet your standards.  Honestly, I wasn't sure if it would be better to have custom export options for overriding, but I didn't think there would be that many different types of exports to make it worth it.  If this code is deemed unworthy to be merged, there will be no hard feelings :)

### DIFF
--- a/src/components/m-table-toolbar.js
+++ b/src/components/m-table-toolbar.js
@@ -74,7 +74,7 @@ export class MTableToolbar extends React.Component {
     doc.setFontSize(15);
     doc.text(this.props.title, 40, 40);
     doc.autoTable(content);
-    doc.save((this.props.exportFileName || this.props.title || 'data') + '.pdf')
+    doc.save((this.props.exportFileName || this.props.title || 'data') + '.pdf');
   }
 
   exportCsv = () => {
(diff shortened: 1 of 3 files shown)


src/components/m-table-toolbar.js
```diff
@@ -53,6 +57,26 @@ export class MTableToolbar extends React.Component {
...
+    const doc = new jsPDF(orientation, unit, size);
+    doc.setFontSize(15);
+    doc.text(this.props.title, 40, 40);
+    doc.autoTable(content);
+    doc.save((this.props.exportFileName || this.props.title || 'data') + '.pdf');
+  }
+
   exportCsv = () => {
     if (this.props.exportCsv) {
       this.props.exportCsv(this.props.columns, this.props.data);
```
mbrn, line 82: Why defaultExportCsv call removed?

mbrn:
Hi @jardon 

Can you check your branch please? Because i couldn't find defaultExportCsv function call.

src/components/m-table-toolbar.js
jardon: No worries.  Its in m-table-toolbar.js on line 49. Let me know if you need anything else!

jardon:
Its definitely there.  I think the git diff is just kind of messed up because I reworked it and reused some old code which makes the diff really confusing.  My apologies.

PR #1998, opened 2020-06-06
Add error state

### LINKED ISSUE #315: Add validation to inline edit.

### LINKED ISSUE #585: Support Validation to record edition
**Is your feature request related to a problem? Please describe.**
The library seems great to me, I'm just starting with it. But I came across a particular problem. Validation of the fields in the insert or update operation.

**Describe the solution you'd like**
I think it would be a plus to be able to define the type of validation for each field or a scheme like Yup, which only works when inserting or updating a record.
If there is an error, you should not confirm the operation and in each field show the current error.

At this moment the validation can be done after confirming the operation and the errors can be shown in another component.

### LINKED ISSUE #602: Displaying Error State for Editable TextFields
Hi, 

Thank you very much for the well-developed component and for your support! Great work! 

I am currently making use of the table's editable feature.
 
When the user edits a cell and clicks on "check" icon button, in onRowUpdate, some validations checks are applied and if the validations fail, all I could do is revert newData with oldData.

However, what I ideally want is to show the user that an error has occurred if the validations have failed.
Option 1: The MaterialUI's TextField error's prop to be set to {true} which will render it as error state (i.e. becomes underlined in red) and not to allow the row update.
Option 2: Revert the data but display an error message somewhere below the footer of the table for example.

I personally prefer Option 1.

Here is what I have tried to do so far but without avail.

          <EditableTable
            enableSearch={false}
            showToolbar={true}
            showTitle={false}
            maxBodyHeight={250}
            title="Coordinates"
            enableRowDelete={true}
            enableRowAdd={true}
            columns={[
              {
                title: 'Count',
                field: 'counter',
                editable: 'onAdd',
                sorting: true,
                type: 'numeric',
                cellStyle: materialTableStyle.centerAlignedCellStyle
              }, {
                title: 'Coordinate',
                field: 'coordinate',
                sorting: false
              }
            ]}
            validationError={this.state.validationError}
            data={this.state.rows}
            editable={{
              onRowAdd: newData => new Promise((resolve, reject) => {
                setTimeout(() => {
                  {
                    this.handleRowAdd(newData, resolve);
                  }
                  resolve()
                }, 1000)
              }),
              onRowUpdate: (newData, oldData) => new Promise((resolve, reject) => {
                setTimeout(() => {
                  {
                    const data = this.state.rows;
                    const index = data.indexOf(oldData);

                    let isValidated = this.validateCoordinates(newData.coordinate);

                    if (!isValidated) {
                      this.setState({validationError: true});
                      return;
                    } else {
                      data[index] = newData;
                      this.setState({
                        data, validationError: false
                      }, () => resolve());
                    }
                  }
                   resolve()
                }, 1000)
              }),
              onRowDelete: oldData => new Promise((resolve, reject) => {
                setTimeout(() => {
                  {
                    this.handleRowDelete(oldData, resolve);
                  }
                  resolve()
                }, 1000)
              })
            }}/>

import React, {Component} from "react";
import {defineMessages, injectIntl} from 'react-intl';
import PropTypes from 'prop-types';
// Material Table imports
import MaterialTable, { MTableToolbar } from 'material-table';
// Material UI
import Paper from '@material-ui/core/Paper';
import TextField from '@material-ui/core/TextField';
// Material UI icons
import AddIcon from '@material-ui/icons/AddBox';
import DeleteIcon from '@material-ui/icons/Delete';
import EditIcon from '@material-ui/icons/Edit';
import CancelIcon from '@material-ui/icons/Save';
import CheckIcon from '@material-ui/icons/Check';
import SearchIcon from '@material-ui/icons/Search';
import ClearIcon from '@material-ui/icons/Clear';

import materialTableStyle, {toolbarStyle} from './EditableTableJSS';
import { withStyles } from '@material-ui/core/styles';

class EditableTable extends Component {

  constructor(props) {
    super(props);
  }

  render() {
    const props = this.props;
    let intl = this.props.intl;
    return(
      <MaterialTable
        title={props.title}
        options={{
          search: props.enableSearch,
          actionsColumnIndex: -1, // aligning table actions on the right-hand side
          toolbar: props.showToolbar,
          showTitle: props.showTitle,
          title: props.title,
          toolbarButtonAlignment: props.toolbarButtonAlignment ? props.toolbarButtonAlignment : 'right',
          doubleHorizontalScroll: false,
          paging: false,
          sorting: true,
          headerStyle: materialTableStyle.headerStyle,
          addRowPosition: 'last',
          maxBodyHeight: props.maxBodyHeight
        }}
        components={{
          // providing class names for the components in order to be able to override their default Material UI CSS
          Container: props => (<Paper {...props} className="material-table-paper" />),
          Toolbar: props => (<StyledMTableToolbar {...props}/>),
          // overriding EditField
          EditField: efProps => {
            return(
            <TextField
              style={efProps.columnDef.type === 'numeric' ? { float: 'right' } : {}}
              type={efProps.columnDef.type === 'numeric' ? 'number' : 'text'}
              placeholder={efProps.columnDef.title}
              value={efProps.value === undefined ? '' : efProps.value}
              onChange={event => efProps.onChange(event.target.value)}
              error={props.validationError}
            />);
          },
        }}
        icons={{
          Add: props => (<AddIcon {...props} color="primary" className="icon-small" />),
          Edit: props => (<EditIcon {...props} className="icon-small" />),
          Search: props => (<SearchIcon {...props} color="primary" className="icon-small" />),
          Delete: props => (<DeleteIcon {...props} className="icon-small" />),
          Check: props => (<CheckIcon {...props} className="icon-small" />),
          Clear: props => (<ClearIcon {...props} className="icon-small" />),
        }}
        localization={{
          body: {
            editTooltip: intl.formatMessage(messages.editTooltip),
            deleteTooltip: intl.formatMessage(messages.deleteTooltip),
            addTooltip: intl.formatMessage(messages.addTooltip),
            emptyDataSourceMessage: intl.formatMessage(messages.noData),
            editRow: {
              deleteText: intl.formatMessage(messages.deleteText),
              cancelTooltip: intl.formatMessage(messages.cancelTooltip),
              saveTooltip: intl.formatMessage(messages.saveTooltip),
            }
          }
        }}
        columns={props.columns}
        data={props.data}
        editable={props.editable}
      />
    );
  }

}

EditableTable.propTypes = {
 /**
 * If true, search field is displayed. *Required
 */
  enableSearch: PropTypes.bool.isRequired,
  /**
  *Display Title
  */
  showTitle: PropTypes.bool,
  /**
  * Title of the table, displayed in the toolbar
  */
  title: PropTypes.string,
  /**
  *By default, toolbar is visible. Set to false to hide the toolbar
  */
  showToolbar: PropTypes.bool,
  /**
  * Alignment by default is right. Can either be right or left
  */
  toolbarButtonAlignment: PropTypes.string,
  /**
  * Allow row deletion. *Required
  */
  enableRowDelete: PropTypes.bool.isRequired,
  /**
  *Allow new row addition. *Required
  */
  enableRowAdd: PropTypes.bool.isRequired,
  /**
  *Table columns. Array. *Required
  */
  columns: PropTypes.array.isRequired,
  /**
  * Table Data/Rows. *Required
  */
  data: PropTypes.array.isRequired,
  /**
  * Editable Data
  */
  editable: PropTypes.object,
  /**
  *Localization
  */
  localization: PropTypes.object,
  /**
  * Max height at which the table becomes scrollable
  */
  maxBodyHeight: PropTypes.number,
  /**
  * Validation Error for all editable TextFields
  */
  validationError: PropTypes.bool,
};

export default injectIntl(EditableTable);

/**
*@function StyledToolbar
* using JSS to apply a className to a component.
*/
function StyledToolbar(props) {
  return (
    <MTableToolbar {...props} />
  );
}
const StyledMTableToolbar = withStyles(toolbarStyle)(StyledToolbar);

This is the result:
image

I can see that the textfield is underlined, however the promise is not resolved and the spinner remains on. 

How can I fix this?

Thanks again!

### LINKED ISSUE #635: Disable save button on add row if all fields aren't complete
Hi,

I was wondering if this functionality exists within the library? I had a look, but I'm not sure it does. I'd like to disable the save button when adding a new row if all the row fields aren't filled out. I hope you can point me in the right direction!

Thank you,

Dhruv

### LINKED ISSUE #744: Issue with Promise.reject()
Hi!
I am trying to add some issue handlers in my table. In remote data Promise I want to use reject, when something will go wrong (according to documentation and tips from other issues).
It's not working correctly tho. I am getting error: _Uncaught (in promise) undefined_. 
In the same place resolve (with proper data ofc) is working great.

Unfortunately it doesn't say much, so I have no clue where to look for the error. 

Have you ever encounter similar issue?

### LINKED ISSUE #1042: How to highlight a wrong field after validate the data ?
First I want to thank you for this contribution to the community that has been very usefu to me.

I was wondering if there is a way to highlight a wrong field (if there is someone) after validate the data when editing a row or adding new one using onRowAdd and onRowUpdate features to show to the user the where is the problem.

I know that i can reject the data using the reject() function of the promise if there are wrong data, but Is there a way to show the user what is the wrong field?  either underlined in red or changing the color to the text of the field ?, is there any function to do that?, because I think that sending an alert or putting a message outside the table does not seem like a very elegant way to solve the problem, the ideal solution would be to highlight the wrong field.

Thank you

### LINKED ISSUE #1152: Validating the TextField when using inline editable
I'm using material table with inline editing mode, and now I want to all of inline editing fields are required, if the user leaves the input is empty or something like that, material table should display error message and disable the buttons.

At the moment, I have to use the `editComponent` to override the input. is there anyway to I can do it by material table without overriding?

### LINKED ISSUE #1576: help - validating field OnRowUpdate, error label is not showing
I have a MaterialTable and I want to validate the fields when editing a line. For example the following code: https://codesandbox.io/s/broken-snow-i4jbi?fontsize=14&hidenavigation=1&theme=dark

I have the function setNameError
`const [nameError, setNameError] = useState({
    error: false,
    label: '',
    helperText: '',
});`

Then the onRowUpdate:
`
onRowUpdate: (newData, oldData) =>
                new Promise((resolve, reject) => {
                    setTimeout(() => {
                        if(newData.name === '') {
                            setNameError({
                                error: true,
                                label: 'required',
                                helperText: 'Required helper text'
                            });
                            reject();
                            return;
                        }
                        resolve();
                        ...
                    }, 600);
                })
`

I want to validate if the field name is empty, if it is empty I want to keep the edition opened and show the error label, like this [image_example](https://i.stack.imgur.com/XGeEt.png)

I can't show the error label, it looks like the setNameError inside the Promise is not working and I don't understand how can I do this.

### LINKED ISSUE #2086: Table close editable mode after reject
**Describe the bug**
When adding a row and reject, the row remains in editable state with data in it but when editing it and reject the editable state of the row closes and old data remains. It happens in version 1.62.0 but not happening in 1.39.0 tested with the same code.

Another problem is that after inserting a char in the field I want to clear the error in the state but the pointer is not in the field after resetting the state so you have to click again in the field.
**To Reproduce**
Try to edit a row and reject in the promise, the editable state will close but the behaviour is different when adding a row.
It is really strange that it happens only if before reject() I set the state of the component but it works fine for add with the same function. I set the state because I have to update the fields with input errors.
```

function Companies({ companies, addCompany, deleteCompany, updateCompany }) {
    const [inputErrors,setInputErrors] = useState(constants.INPUT_ERRORS);
    //Commented until bug is fixed on mui table
    // const setErrorForField = (fieldName, value) =>{
    //     if(value != inputErrors[fieldName]){
    //         const errors = {...inputErrors}
    //         errors[fieldName] = value
    //         setInputErrors(errors);
    //     }
    // }

    const generateTextField = (label, type, props, field) => {
        return <TextField 
                type={type}
                error={inputErrors[field]}
                label={label}
                onChange={(e) => {
                    //commented because after inserting one char in field state is changed and textfield not focuse anymore, so no writing in the field after first char
                    //setErrorForField(field, false);
                    props.onChange(e.target.value);
                }}
                value={props.value === undefined ? '' : props.value}
            />
    }

    const actionFunction = (callback, newData, isUpdate) => new Promise((resolve,reject) => {
        setTimeout(() => {
            let ok = true;
            let errors = {...constants.INPUT_ERRORS}

            const onError = (fieldKey) =>{
                if(!newData[fieldKey]){
                    errors[fieldKey]= true;
                    return false;
                }
                return ok;
            }

            if(!newData[constants.EMAIL] || !EmailValidator.validate(newData[constants.EMAIL])){
                errors[constants.EMAIL]= true;
                ok=false;
            }

            let phoneNumber = null;
            if(newData[constants.PHONE_NUMBER])
                phoneNumber=parsePhoneNumberFromString(newData[constants.PHONE_NUMBER], 'RO');
            
            if(!newData[constants.PHONE_NUMBER] || !phoneNumber || !phoneNumber.isValid() ){
                errors[constants.PHONE_NUMBER]= true;
                ok=false;
            }else{
                newData[constants.PHONE_NUMBER] = phoneNumber.formatInternational();
            }

            ok = onError(constants.NAME);
            ok = onError(constants.CUI);
            ok = onError(constants.DOMAIN);                        

            if(ok){
                resolve();
                callback(newData);
                errors = {...constants.INPUT_ERRORS};
            }
            else{
                reject();
            }
               //this line 
                setInputErrors(errors);
        }, 600);
    })

    const columns = [
        { title: 'Name', field: constants.NAME,
        editComponent: (props) => generateTextField("Name", "Name",props, constants.NAME)},
        { title: 'CUI', field: constants.CUI,
        editComponent: (props) => generateTextField("CUI", "",props,constants.CUI)},
        { title: 'Email', field: constants.EMAIL,
        editComponent: (props) => generateTextField("Email", "email",props,constants.EMAIL)},
        {
            title: 'Phone Number',
            field: constants.PHONE_NUMBER,
            editComponent: (props) => generateTextField("Phone Number", "tel",props,constants.PHONE_NUMBER)
        },
        { 
            title: 'Domain', 
            field: constants.DOMAIN,            
            editComponent: (props) => generateTextField("Domain", "",props,constants.DOMAIN)
        }
    ]

    return (
        <MaterialTable
            icons={tableIcons}
            title="Vizualizare firme"
            columns={columns}
            data={companies}
            editable={{
                onRowAdd: (newData) => actionFunction(addCompany, newData),
                onRowAddCancelled: () => setInputErrors(constants.INPUT_ERRORS),
                onRowUpdateCancelled: () => setInputErrors(constants.INPUT_ERRORS),
                onRowUpdate: (newData/*, oldData*/) =>  actionFunction(updateCompany, newData, true),
                onRowDelete: (oldData) =>
                    new Promise((resolve) => {
                        setTimeout(() => {
                            resolve();
                            deleteCompany(oldData.ID);
                        }, 600);
                    }),
            }}
        />
    );
}
```
**Expected behavior**
After rejecting the row should remain in editable state.
Same behaviour as for add row.

**Additional context**
    "material-table": "1.62.0",
     React

Domino987:
## Related Issue
Fix #744 Fix #1576 Fix #1152 Fix #1042 Fix #635 Fix #602 Fix #585 Fix #315 Fix #2086
## Description
This PR add an error state to the table.
If the data promise rejects or fails, the loading state will not be shown indefinitely, but instead an error message will be shown and a button to retry the last query, to try to fix the error.

If onAdd/onEdit/onDelete fails, the reject will be caught and put into the error state as well. This will be passed to render, so the user can decide on how to show that error.

In this context, a validate prop is added to the colum. It is a function and either return a boolean (for the red bar below), a string (for red bar and error text) and an object: { isValid: boolean, helperText?: string }. This validate can be done by the user e.g with yup or manually.

Let me know what you think!

Once this is done, a section should be added to the docs to explain these validation process.

* prop-types
* material-table
* index / types
* default props

## Additional Notes
image


j3den:
Thank you for working on this!

I am not a JS dev (I am a java/spring dev so my understanding is shakey at best)...

I think I can see you set the loading state to false and you allow some form of component to be displayed with a 'retry' icon?

I like the idea of the retry option. Just not sure what the new ```overlayError``` prop would render/cause?

Is there a possibility to add custom error handling...

Something along these lines?
```
<MaterialTable 
onRemoteError={(err)=>console.error("handle error")},
/>

```
Again thansk for the PR :).

Domino987:
> Thank you for working on this!
> 
> I am not a JS dev (I am a java/spring dev so my understanding is shakey at best)...
> 
> I think I can see you set the loading state to false and you allow some form of component to be displayed with a 'retry' icon?
> 
> I like the idea of the retry option. Just not sure what the new `overlayError` prop would render/cause?
> 
> Is there a possibility to add custom error handling...
> 
> Something along these lines?
> 
> ```
> <MaterialTable 
> onRemoteError={(err)=>console.error("handle error")},
> />
> ```
> 
> Again thansk for the PR :).

As seen in the added image, the overlayError renders this error text with the retry button. This will allow the user to retry the action.

Currently, only the data promise is supported, but in the future, the edit/add/delete calls could be added to the error state as well. 

How would you define the onRemoteError callback? Would you return an error text, pass the error up to an error boundary or what do you have in mind?

j3den:
> How would you define the onRemoteError callback? Would you return an error text, pass the error up to an error boundary or what do you have in mind?

Was thinking it would be the error passed in the ```reject(error)``` promise declared in the data/query prop? So bubbling the rejection up I guess?

Currently, the catch() is unhandled?
Screenshot from 2020-06-08 15-14-42

Screenshot from 2020-06-08 15-16-55

Maybe the new PR will mean the catch() block can handle some custom logic meaning the extra bit I requested will be redundant?

Domino987:
It will catch the reject and show the error message. This callback could maybe be used to restore the table like a reset, if the default behavior (reload the current query) is not desired.
I will think about it.

darewreck54:
Couple options
1) Have reject take in an object which gets propagated upwards to the render method for edit.  This edit method would have an addition field which it can grab the rejected message
2) Add validation for the row.  This would require you to add some new property for the component that takes in validation for the row.  The validation gets called every time the row is mutated and the error gets propagated to the render method so user can do what the want.

src/default-props.js
```diff
@@ -18,6 +18,21 @@ OverlayLoading.propTypes = {
   theme: PropTypes.any
 };
 
+const OverlayError = props => (
+  <div style={{ display: 'table', width: '100%', height: '100%', backgroundColor: fade(props.theme.palette.background.paper, 0.7) }}>
+    <div style={{ display: 'table-cell', width: '100%', height: '100%', verticalAlign: 'middle', textAlign: 'center' }}>
```
darewreck54, line 23: nit: Be nice to abstract styling away from the html by having a js object that represent the style and pass it into this div.

darewreck54:
@Domino987 any idea when this will be out? or if any of the advice with propagating the promise reject object to render?

darewreck54:
Also the example you have shows what happens when you try to load data?  How about when you input data with the save.  Any example of how that would look like?

Domino987:
> Also the example you have shows what happens when you try to load data? How about when you input data with the save. Any example of how that would look like?

Not yet, but I am passing the error object to the editComponent, so if an error happens, the user could handle it.



> 2\. Add validation for the row.  This would require you to add some new property for the component that takes in validation for the row.  The validation gets called every time the row is mutated and the error gets propagated to the render method so user can do what the want.

I added a 'validate' object to the column, to show an error text with error state.
It accepts a boolean (for the red bar below), a string (for red bar and error text) and an object: { isValid: boolean, helperText?: string }. This validate can be done by the user e.g with yup.

Domino987:
> any idea when this will be out

@darewreck54 No idea yet, I will see what @mbrn thinks about it and it needs quite a lot of testing. So if you want to try it out, download this branch and give it a go to see if that works for you.

src/default-props.js
Domino987: Thats the way mbrn decided tostyle the overlays. Where would you pass that style object for each div?

patrickp-dev:
Hey @Domino987, so i inadvertently also made a similar fix,  here's the [pull request](https://github.com/mbrn/material-table/pull/2044).

It looks as though you're accomodating for some additional cases that I haven't (add, update, delete errors).

Some feedback on the implementation:

- I'm not sure I agree with the overlay as the means of displaying the error message. In its current state, it forces the table into a state where the only action that can occur is a reload. This means for things such as update, add and delete, a refresh is required, rather than a retry of the action (click the save button again to resubmit for add/update, click the delete icon again for delete).
- Material design doesn't really provide any guidance on how to handle errors within the table, so we'll need to make a decision that best aligns with the existing design system. In my opinion, I think the error should be relative to the element that caused it. 
  - For table level errors such as the query error, i think displaying an error message row better fits as it's inline with the existing behaviour of displaying a message when there are no rows and doesn't prevent access to other table functionality such as modifying filters or selecting table actions. This is especially applicable if it is the current table state (filters or search value) that is causing the error as the current implementation doesn't allow the user to recover from the error state. 
  - For row level errors (add, remove, update), i feel like we should be displaying the error message inline. I had a thought that maybe we could have the error message partially overlay within the table row, kind of like a dismissable snackbar. Below is an example i quickly mocked up
image

The above ideas are up for discussion. Just some suggestions to take into consideration.

Domino987:
> Hey @Domino987, so i inadvertently also made a similar fix, here's the [pull request](https://github.com/mbrn/material-table/pull/2044).
> 
> It looks as though you're accomodating for some additional cases that I haven't (add, update, delete errors).
> 
> Some feedback on the implementation:
> 
> * I'm not sure I agree with the overlay as the means of displaying the error message. In its current state, it forces the table into a state where the only action that can occur is a reload. This means for things such as update, add and delete, a refresh is required, rather than a retry of the action (click the save button again to resubmit for add/update, click the delete icon again for delete).
> * Material design doesn't really provide any guidance on how to handle errors within the table, so we'll need to make a decision that best aligns with the existing design system. In my opinion, I think the error should be relative to the element that caused it.
>   
>   * For table level errors such as the query error, i think displaying an error message row better fits as it's inline with the existing behaviour of displaying a message when there are no rows and doesn't prevent access to other table functionality such as modifying filters or selecting table actions. This is especially applicable if it is the current table state (filters or search value) that is causing the error as the current implementation doesn't allow the user to recover from the error state.
>   * For row level errors (add, remove, update), i feel like we should be displaying the error message inline. I had a thought that maybe we could have the error message partially overlay within the table row, kind of like a dismissable snackbar. Below is an example i quickly mocked up
>     image
> 
> The above ideas are up for discussion. Just some suggestions to take into consideration.

Hi thanks for your feedback.

The overlay will only be displayed for the query fail, which will trigger a new query. Since errorState will be passed to the query as well, the user can decide maybe to reset the table on query, if an error is provided or how he wants to handle that error on retry.
Update/add/Delete will only be passed to the editComponent/ editCell and the user has to decide on how to render those erros.
I was also thinking about using maybe the snackbar component to render the errors non-blocking but thats up to debate.

What do you think?

tschaka1904:
Looking forward to have the validate functionality in place! 👍

hungbang:
pretty good. thanks for your work @Domino987

Domino987:
@tschaka1904 can you check if the YUP validation works for that too.

tschaka1904:
Will do @Domino987

tschaka1904:
Alright! I guess got it working. I took the projects `index.html`/`demo.js` to get it working. 

Basically, into demo js put those two: 

```
const schema =
    Yup.object().shape({
        name: Yup.string().required(),
        surname: Yup.string().required(),
        [....]
    });


const validate = (values) => {
  try {
    schema.validateSync(values, {stripUnknown: true});
    return {isValid: true, helperText: ''}
  } catch (e) {
    return {isValid: false, helperText: 'eMessage'}
  }
};
```

Plus into the I extended the `columns` array by this: `validate: validate`. 

I replaced the `switch` statement in `m-table-edit-row.js` with:

```
if (columnDef.validate) {
    error = columnDef.validate(this.state.data);
}
```

All of that together resulted in: 



All that looking pretty nice! 👍 

Some thought on the `switch/case`: 
Whilst I can see what the general idea is, I think, eternalising the actual valid/invalid logic would be a good idea. This way it becomes more flexible, and you would just need to ensure that you provide `{isValid: true, helperText: 'yourMsg'` is suiting your needs. But maybe I also completely missed the point there. 

Also, there are still a few warning appearing in my console. Like: 
- `Warning: Failed prop type: Invalid prop `errorState` of type `boolean` supplied to `MTableCell`, expected `object`.`

or 

- `Warning: React does not recognize the `helperText` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `helpertext` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`


Thanks for the work @Domino987! 🥇  literally the right feature at the right time for me!

Domino987:
> Also, there are still a few warning messages appearing in my console. Like:

I updated the PR to remove these warnings by removing the passing to the input fields. So that's done.

I wanted to make it as easy possible for the users. So example if somebody does not use YUP, you could just do:
`valdiate: rowData => rowDate.name.length > 2`
`valdiate: rowData => rowDate.name.length > 2 ? "Name to short" : true`

This would make it really easy to use it if YUP is not wanted eg.

tschaka1904:
> I wanted to make it as easy possible for the users. So example if somebody does not use YUP, you could just do: [...]

@Domino987, yeah, understood! 👍 

Looking forward seeing this in master, @mbrn! :)

types/index.d.ts
```diff
@@ -99,6 +104,7 @@ export interface EditComponentProps<RowData extends object> {
   onChange: (newValue: any) => void;
   onRowDataChange: (newValue: RowData) => void;
   columnDef: EditCellColumnDef;
+  errorState?: ErrorStateL
```
tschaka1904, line 107: I assume the L shouldn't be there?

Domino987: Thanks.

Domino987:
@tschaka1904 We of course have to update the docs, too with examples. Once this is through we need to add some examples.

src/components/m-table-edit-row.js
```diff
@@ -67,8 +67,8 @@ export default class MTableEditRow extends React.Component {
         if (columnDef.editable === 'onUpdate' && this.props.mode === 'update') {
           allowEditing = true;
         }
-        if (typeof columnDef.editable == 'function'){
-            allowEditing = columnDef.editable(columnDef, this.props.data);
+        if (typeof columnDef.editable == 'function') {
```
IniterWorker, line 70: Should be replaced by the triple equals operator.

IniterWorker, line 70:
```suggestion
        if (typeof columnDef.editable === 'function') {
```

Domino987: Obviously, thanks:)

darewreck54:
Thanks for the work.  I just tested it out and it worked.   One comment is that it's not really intuitive that if you return an error that is not true like a string, that it translate to helperText in the editComponent.  Some of the other libraries i've used like formik would just return undefined meaning that there is no error and if there is text, then you know it's an error.

darewreck54:
```

export interface EditComponentProps<RowData extends object> {
  rowData: RowData;
  value: any;
  onChange: (newValue: any) => void;
  onRowDataChange: (newValue: RowData) => void;
  columnDef: EditCellColumnDef;
  errorState?: ErrorState;
}
```

Does the interface need to be updated?   I'm not seeing an error in the typescript interface, but I do see the error and helperfield when i run the code.  I'm using     "material-table": "1.64.0",

Domino987:
@derek thanks for your response. I must have missed that ts def, can you create a PR?

darewreck54:
@Domino987 I notice that even thou validate fails, you still can save it.  It seems redundant that you would have to implement it in the validate and in the edit onAdd methods.  Or am i doing it incorrectly?

darewreck54:
> @derek thanks for your response. I must have missed that ts def, can you create a PR?

I'll see what I can do.  I'm not really familiar  with the code base.  If you can point me to the files, that would be helpful.

darewreck54:
image

Yeah without the updated typescript def i think you get this error

IniterWorker:
I have tested with Yup.

I also read the @tschaka1904 way to handle it.
- A little issue if we use to validate as a general validation handler, every field will raise an error field text.

I tried to write something more flexible.

```js
export const validator = schema => values => {
    try {
        schema.validateSync(values, { stripUnknown: true });
        return { isValid: true, helperText: '' };
    } catch (e) {
        return { isValid: false, helperText: e.message };
    }
};
```

```js
{
      title: t('my_field_translation_key'),
      field: 'my_field',
      validate: validator(
          Yup.object().shape({
                my_field: Yup.string()
                        .min(1)
                        .max(24)
                        .required()
                 })
            )
},
```

But, I'm currently wondering if we could do something to add a full Yup support easily. Or, maybe, we could find a functional alternative.

B. r

Domino987:
> @Domino987 I notice that even thou validate fails, you still can save it. It seems redundant that you would have to implement it in the validate and in the edit onAdd methods. Or am i doing it incorrectly?
> 
> ```
>  onRowAdd: (newData) =>
>           new Promise((resolve) => {
>             setTimeout(() => {
>               resolve();
>               const rowData = props.value ? [...props.value] : [];
>               rowData.push(newData);
>               if (props.onChange) {
>                 props.onChange(rowData);
>               }
>             }, ANIMICATION_TIMEOUT_IN_MS);
>           }),
> ```
> 
> How do you actually know that an error occurred from the validate method? Or is there a way to block it from onRowAdd?

It should block the onRowAdd call, if the valdiation fails. Can you create a sandbox to reproduce that behavior? The onRowAdd call should only be invoked, if the validation goes through.

Domino987:
> Yeah without the updated typescript def i think you get this error

This is not connected to the interface but to a incorrect passing. Can you put that in a sandbox as well?

Domino987:
> I have tested with Yup.
> 
> I also read the @tschaka1904 way to handle it.
> 
> * A little issue if we use to validate as a general validation handler, every field will raise an error field text.
> 
> I tried to write something more flexible.
> 
> ```js
> export const validator = schema => values => {
>     try {
>         schema.validateSync(values, { stripUnknown: true });
>         return { isValid: true, helperText: '' };
>     } catch (e) {
>         return { isValid: false, helperText: e.message };
>     }
> };
> ```
> 
> ```js
> {
>       title: t('my_field_translation_key'),
>       field: 'my_field',
>       validate: validator(
>           Yup.object().shape({
>                 my_field: Yup.string()
>                         .min(1)
>                         .max(24)
>                         .required()
>                  })
>             )
> },
> ```
> 
> But, I'm currently wondering if we could do something to add a full Yup support easily. Or, maybe, we could find a functional alternative.
> 
> B. r

Hi thank you for your response.
I wanted it to be as flexibel as possible so that users are not tied to YUP but they can do it however they want. That's why you can use anything for the validation (also YUP for each cell). YUP support could be added by adding a global validator schema to the table and use that instead. That would dramatically increase the complexity and I do not know if that's the way to go. I will consider it.

darewreck54:
> > @Domino987 I notice that even thou validate fails, you still can save it. It seems redundant that you would have to implement it in the validate and in the edit onAdd methods. Or am i doing it incorrectly?
> > ```
> >  onRowAdd: (newData) =>
> >           new Promise((resolve) => {
> >             setTimeout(() => {
> >               resolve();
> >               const rowData = props.value ? [...props.value] : [];
> >               rowData.push(newData);
> >               if (props.onChange) {
> >                 props.onChange(rowData);
> >               }
> >             }, ANIMICATION_TIMEOUT_IN_MS);
> >           }),
> > ```
> > 
> > 
> > How do you actually know that an error occurred from the validate method? Or is there a way to block it from onRowAdd?
> 
> It should block the onRowAdd call, if the valdiation fails. Can you create a sandbox to reproduce that behavior? The onRowAdd call should only be invoked, if the validation goes through.

I looked into it deeper.  Basically, if you don't return false it will won't block.  In my case, I'm passing a string so that I'm able to display the error message in the render edit component.

```
function Editable(data: any, setData: any) {
  const columns = [
    { title: 'Surname', field: 'surname', initialEditValue: 'initial edit value' },
    {
      title: 'Birth Place',
      field: 'birthCity',
      validate: (value: any) => {
        if(value.name === 'test') {
          return "Invalid";
        } else {
          return true;
        }
      },
    },
  ];

  return (
    <MaterialTable
      icons={tableIcons as any}
      title="Editable Preview"
      columns={columns as any}
      data={data}
      components={{
        Container: (componentProps) => {
          return <Paper elevation={2} {...componentProps} />
        },
      }}
      editable={{
        onRowAdd: newData =>
          new Promise((resolve, reject) => {
            setTimeout(() => {
              setData([...data, newData]);
              
              resolve();
            }, 1000)
          }),
        onRowUpdate: (newData, oldData) =>
          new Promise((resolve, reject) => {
            setTimeout(() => {
              const dataUpdate = [...data];
              if(oldData && (oldData as any).tableData) {
                const index = ((oldData as any).tableData as any).id;
                dataUpdate[index] = newData;
                setData([...dataUpdate]);
                resolve();
              } else {
                reject();
              }
             
            }, 1000)
          }),
        onRowDelete: oldData =>
          new Promise((resolve, reject) => {
            setTimeout(() => {
              const dataDelete = [...data];
              const index = (oldData as any).tableData.id;
              dataDelete.splice(index, 1);
              setData([...dataDelete]);
              
              resolve()
            }, 1000)
          }),
      }}
    />
  )
}
```

image


vs 

image

Domino987:
Good catch , I will fix that . Thank you for the detailed response.

darewreck54:
> image
> 
> Yeah without the updated typescript def i think you get this error

I think you get this error when just using the simple edit example as well.

Domino987:
> looked into it deeper. Basically, if you don't return false it will won't block. In my case, I'm passing a string so that I'm able to display the error message in the render edit component.

I updated the PR to show an error for the select component.

darewreck54:
@Domino987 I had another question for you.  Right now the validate only takes in the value.  Does it make sense to take it the whole value of the row (entry).

The use case: 
Say that a row of the table is:
{
   id:string,
   name: string
}

Let's say you i want to validate that there is no duplicate names.  So right now with the validate interface you can't really achieve that through an update method.  If you include the whole row, then I can compare the ID. 

Let me know what your thoughts on this are?
Thanks,
Derek

Domino987:
@darewreck54 It passes the whole row:
```{
birthCity:63
birthYear:1987
name:"Mehmet"
surname:"Baran"
tableData:{editing: "update", id: 0}
}
```
Do you mean all data passed into the table, so all rows?

darewreck54:
Let me look into this since I’m not seeing the whole two data.  Maybe I’m missing something

Sent from my iPhone

On Jul 11, 2020, at 12:13 AM, Domino987 <notifications@github.com> wrote:

﻿

@darewreck54<https://github.com/darewreck54> It passes the whole row:

birthCity:63
birthYear:1987
name:"Mehmet"
surname:"Baran"
tableData:{editing: "update", id: 0}
}


Do you mean all data passed into the table, so all rows?

Domino987:
Ok maybe we are misunderstanding each other. We have several rows, but you only can validate one row at a time. So you cannot currently access all your rows in the table.

darewreck54:
Yeah I wanted to see the whole row but I’m only seeing the col of that row being passed into the validate. But let me look at
My example and see if I have configured it incorrectly. I don’t even see the table index being passed in.

Sent from my iPhone

On Jul 11, 2020, at 9:16 AM, Domino987 <notifications@github.com> wrote:

﻿

Ok maybe we are misunderstanding each other. We have several rows, but you only can validate one row at a time. So you cannot currently access all your rows in the table.

darewreck54:
@Domino987  Checkout this example: https://codesandbox.io/s/material-table-starter-template-ik4v9

image


I would expect to see more then just the birthday printed out.  I would expect to see the table Index plus the other info for the row.

Domino987:
@darewreck54 
I get the object by just opening your sandbox.
image

darewreck54:
@Domino987 Thanks, sorry didn't realized it was create.  

> > looked into it deeper. Basically, if you don't return false it will won't block. In my case, I'm passing a string so that I'm able to display the error message in the render edit component.
> 
> I updated the PR to show an error for the select component.

Was this PR fix every released?  I'm using the newest one and it still not fixed.  https://codesandbox.io/s/material-table-starter-template-ik4v9?file=/src/index.js

hamiltondanielb:
I don't see the loader after a reject, but I do NOT see the error message, just the data that was there before

```           
 data={(query: any) => {
              const qs = switchPage(query.page);
              const filter = applyFilters(query);
              return new Promise((resolve, reject) => {
                getTechicalAssessments(query.pageSize, filter, qs)
                  .then(technicalAssessmmentsData => {
                    resolve({
                      data: technicalAssessmmentsData.data,
                      page: query.page,
                      totalCount: technicalAssessmmentsData.originalTotalCount,
                    });
                  })
                  .catch(err => {
                    console.log(err);
                    reject({ message: "fix me" });
                  });
              });
            }}
```

PR #2007, opened 2020-06-08
Fix initialPage value not being applied when using a remote data sour…

patrickp-dev:
…ce. Add material-icon as dev dependency. Move demo elements into expansion panels.

## Related Issue
#964

## Description
Fix for the initialPage prop not being applied when using remote data sources.

## Key Changes

- setting paging to false when data source is remote as other parameters such as search, filter and sort delegate the responsibility to the data source rather than the table. This assumption should apply to the paging as well as it's already expecting the count and page to be returned by the data source.
- setting the isLoading to true when the state is initially constructed. On mount it is set to false if the data source is not remote, else left as true until the data source returns.
- Page overflow check in the onComponentUpdate is no longer applied while isLoading is true (prevents infinite update loop due to page being > 0 and count === 0 on mount)

## Additional Changes
I can revert these changes if preferred.
- Added expansion panels to the demo page. It made it easier to test the on mount logic as the tables remount when the panel is expanded and removes the need to scroll to the remote data table on refresh.
- Added the material icon library as a dev dependency. Only needed it for the expansion panel icon.

## Related PRs
#974
#1518
#1381

- Remote data sources
- Static data sources
This may introduce breaking changes to dependents that expect the table to apply it's pagination/result set trimming when using remote data sources that return a result set greater than the page size. With this change the pagination/result set trimming is no longer applied by the table. If the result set is now greater than the page size, a result set > the page size will be displayed. I specifically have decided not to address this as in my opinion if the table is delegating search etc. to the remote data source, it is the data source's responsibility to trim the result set to size. If it doesn't, it's a bug in the data source, not the table.

### Remote Data source
- [ ] When intialPage is set, query object passed to data source contains the initialPage value.
- [ ] When intialPage is unset, query object passed to data source contains the default value of 0.
- [ ] When initialPage is greater than count returned by the data source (pageSize * page > count), the page current page is set to the max page.

### Static Data Source
- [ ] When intialPage is set, pagination should be set to the initial value on load.
- [ ] When intialPage is unset, the initial value should be 0.
- [ ] When initialPage is greater than count of the data source, the page should be set to the max page value.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -1,9 +1,13 @@
 import { Grid, MuiThemeProvider, Button } from '@material-ui/core';
 import { createMuiTheme } from '@material-ui/core/styles';
+import ExpansionPanel from '@material-ui/core/ExpansionPanel';
+import ExpansionPanelSummary from '@material-ui/core/ExpansionPanelSummary';
+import ExpansionPanelDetails from '@material-ui/core/ExpansionPanelDetails';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import Typography from '@material-ui/core/Typography';
 import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
 import MaterialTable from '../src';
-import Typography from "@material-ui/core/Typography";
 
 let direction = 'ltr';
 // direction = 'rtl';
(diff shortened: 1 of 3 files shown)


mbrn:
Please resolve conflicts

mbrn:
I will close PR. You can reopen it after resolving all conflicts.

shivamragnar:
This should have been released...this is a bug with this cool library.

Issue #1429, opened 2019-12-09
endless loading/spinning when using function for data (remote data)


hbinkle:
the table rerenders/spinns endless when using a fuction for data instaed of a preloaded object.
- Not matter if I call resolve or reject.

https://codesandbox.io/s/material-demo-yvqym?fontsize=14&hidenavigation=1&theme=dark

SvenNussgruber:
In your sandbox the query string is the URL to your sandbox. Did you mean to query an API endpoint instead?

hbinkle:
> In your sandbox the query string is the URL to your sandbox. Did you mean to query an API endpoint instead?

Right, it should be an API endpoint when the code is really used. But in the demo case it does't matter because the response is not used. I could also have used "http://www.google.com". It simply needs to be a successfull fetch to demonstrate.

pjshu:
same problem

cinnabarcaracal:
We're getting this same issue in our project. The table is given a function that returns a promise, and it can fetch data normally just fine. But if the promise fails, the table `isLoading` state stays as `true`.

I can tell the promise is failing because it will execute a `.catch` if I put one in there

cinnabarcaracal:
I think the problem is in the `onQueryChange` function in `material-table.js`.

Potential solution:
```
onQueryChange = (query, callback) => {
  query = { ...this.state.query, ...query };

  this.setState({ isLoading: true }, () => {
    this.props
      .data(query)
      .then(result => {
        query.totalCount = result.totalCount;
        query.page = result.page;
        this.dataManager.setData(result.data);
        this.setState(
          {
            isLoading: false,
            ...this.dataManager.getRenderState(),
            query,
          },
          () => {
            callback && callback();
          },
        );
      })
      .catch(reason => {
        this.setState({ isLoading: false });
      });
  });
};
```

restaadiputra:
Is there any news about this? I am having the same issue and as for my current solution is accessing the table ref in my .catch and call the setState manually to stop the loading animation, 

`
.catch(error => {
        tableRef.current.setState({
          isLoading: false
        });
        console.log('ERROR ', error);
});
`

cinnabarcaracal:
Please don't close this issue

PR #2044, opened 2020-06-14
Update table to track errors in the remote data source and display a …

patrickp-dev:
…message when they occur.

## Related Issue
#1429 

## Description
Add error handling for remote data sources.

This change only accounts for error handling with remote data sources. It's assumed that static data sources can't fail as they are defined/generated external to the table.

Introduced `isDataSourceError` state for tracking if an error has occurred in the data source.
Add `error` property to `query` state and pass it through to the table body. It is not currently used however allows for components overriding the MTableBody to have access to the error if required.

Add `showEmptyDataSourceMessage` option to control whether the error row is displayed.
Add `localization. dataSourceErrorMessage ` translation for the error row message.

* onQueryChange
* Table body

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -195,21 +195,21 @@ class App extends Component {
                   searchPlaceholder: "Outlined Search Field",
                 }
               }}
-              data={query => new Promise((resolve, reject) => {
+              data={query => {
                 let url = 'https://reqres.in/api/users?'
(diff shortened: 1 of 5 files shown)


mbrn:
Please resolve conflicts

patrickp-dev:
Closing as issue resolved by change in https://github.com/mbrn/material-table/pull/1998

PR #2070, opened 2020-06-18
adds a new table option which allows a user to pass properties to the…

justGunnar-leaseup:
… select all checkbox in the table header

## Related Issue
I didn't open one, figured it would be faster to just PR.

## Description
selectionProps are not applied to the select all checkbox in the header. Because we want to allow users to pass a function based on row data for selectionProps, and the header does not have rowData, we need a second parameter specifically for the select all checkbox. 

## Related PRs
None

*m-table-header, prop-types, default-props, m-table

## Additional Notes
Please let me know if there's anything else needed here. I'd love to use this library moving forward. Its clean, easy to configure, and easy to understand.


mbrn:
Please resolve conflicts

justGunnar-leaseup:
hey @mbrn is that prettier failure something i can fix? i see another PR with the same error. The fork/branch exist so id expect a git checkout to work fine.

Issue #1785, opened 2020-04-01
Wrong iteration logic in 'onSelectionChange()'


ifndefdeadmau5:
**Describe the bug**
As title already states, the logic in onSelectionChange is bit wrong.

**To Reproduce**
1. Use 'selection: true' option 
2. Provide dataset where each data has bit different property(In normal dataset, I couldn't reproduce this error)

**Expected behavior**
'rows' parameter passed to onSelectionChange should not contain duplicate items

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Issue #322, opened 2019-03-14
onSelectionChange does not fire for child rows, in tree view.


shabeelsyed:
**Describe the bug**
onSelectionChange does not fire for child rows, in tree view.

**To Reproduce**
https://mbrn.github.io/material-table/#/docz-examples-13-example-tree-data
Debug the code, it fires only for top most rows, not for its children.
onSelectionChange = () => {
    if (this.props.onSelectionChange) {
      const selectedRows = this.state.data.filter(row => row.tableData.checked);
      this.props.onSelectionChange(selectedRows);
    }
  }

**Expected behavior**
Fire it for all the rows including parent and child.

mbrn:
This is done with 1.27.1

shabeelsyed:
Thanks, I verified!

PR #2085, opened 2020-06-24
Fix redundant recursive check in onSelectionChange

ifndefdeadmau5:
## Related Issue

#1785 
#322

## Description

when both `selection` and `parentChildData` are used together, `onSelectionChange` callback receives duplicate rows, because there's a logic inspecting `tableData.childRows` while iterating rows, but since the function is already iterating every rows including child rows, there's no need to check child rows twice in this case.

I've only tested with 1 level depth tree data, so not sure if there are other corner case

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -438,8 +438,6 @@ export default class MaterialTable extends React.Component {
           if (row.tableData.checked) {
             selectedRows.push(row);
           }
-
-          row.tableData.childRows && findSelecteds(row.tableData.childRows);
         });
       };
 


src/material-table.js
```diff
@@ -438,8 +438,6 @@ export default class MaterialTable extends React.Component {
           if (row.tableData.checked) {
             selectedRows.push(row);
           }
-
-          row.tableData.childRows && findSelecteds(row.tableData.childRows);
```
mbrn, line 442: Are you sure removing this line? Because if data has treefied it will find selected data as nestedly.

PR #2097, opened 2020-06-25
Changes ISO dates to use date-fns's parseISO function

### LINKED ISSUE #2079: MaterialTable with column type "date" is off by one
# Guidelines

**To Reproduce**

```
const columns = [
            {
                title: "foo",
                field: "date",
                headerStyle: {textTransform: "capitalize"},
                type: "date",
                sorting: false
            },
            {title: "Views", field: "visits", type: "numeric"},
            {
                title: "Time Spent",
                field: "timeSpent",
                type: "numeric",
                render: (row) => prettyMilliseconds(row.timeSpent)
            }
        ];
data = [
            {
                date: '2020-06-13',
                timeSpent: 3,
                visits: 2,
            },
        ];

<MaterialTable
          columns={columns}
          data={data}
          title={title}
          icons={tableIcons}
        />
```

**Expected behavior**
The table should have one row with date 6/13/2020

**Screenshots**
image

**Workaround**
I can get the correct date to show by removing the line `type: 'date'` which gives me 2020-06-13 which isn't aesthetic but is correct.

Latest Chrome MacOS Mojave

material-table 1.62.0

ajmueller:
Fixes #2079

## Related Issue
Relate the Github issue with this PR using `#2079`

## Description
Changes ISO dates to use date-fns's parseISO function instead of Date constructor.

* Parsing data for columns with types of `date`, `time`, or `datetime`.


src/components/m-table-cell.js
```diff
@@ -36,23 +37,23 @@ export default class MTableCell extends React.Component {
       if (this.props.value instanceof Date) {
         return this.props.value.toLocaleDateString();
       } else if(isoDateRegex.exec(this.props.value)) {
```
Domino987, line 39: Can you also add dateLocale here as seen here: parseISO(this.props.value).toLocaleTimeString(dateLocale);

mbrn:
Please resolve conflicts

src/components/m-table-cell.js
ajmueller: @Domino987 assuming you meant on line 40, I've just pushed that change.

ajmueller:
@Domino987 @mbrn sorry for the delay; I've been out of town the last week and a half.

I've pushed the requested change and resolved conflicts.  Let me know if there's anything else!

Domino987:
Hi thanks for the change. Looking good. I am sure, it will be merged soon. We still have some problems with the prettier cli . It should automatically prettify your code on commit. I think it's because the pr was created before the prettier was added.

PR #2102, opened 2020-06-27
Allow exportFileName  to be a function

### LINKED ISSUE #1806: unable to call defaultExportCsv() from custom overridden exportCsv() function
I want to have a timestamp as the file name, passing the file name as property to options.exportFileName will get the timestamp when the page got rendered instead of when the export button was clicked. Hence I was trying something like below and I'm unable to call the `defaultExportCsv()` from the overridden `exportCsv()`

```
function Index({model}) {
  const handleExportCsv = () => {
    // set filename as timestamp here
    // call defaultExportCsv() here 
  }

  return (
    <MaterialTable
      columns={[
        { title: 'ID', field: 'id' },
      ]}
      data={model}
      title="Title"
      options={{
        columnsButton: true,
        exportButton: true,
        actionsColumnIndex: -1,
        exportCsv: handleExportCsv
      }}
    />
  );
}
```

I have opened [stackoverflow question](https://stackoverflow.com/questions/61079702/how-to-call-defaultexportcsv-of-reactjs-material-table-from-the-overriden-expo) as well but didn't get answers there yet.

Domino987:
## Related Issue
fixes #1806

## Description
This allows exportFileName to be function to make the export name adapt to get the current date or to respond to the current data/columns passed..


mbrn:
Please resolve conflicts

Issue #2112, opened 2020-06-30
'WebpackError: ReferenceError: window is not defined' when building a static site using gatsbyjs


floAr:
Using the current version of material-table it is not possible to build out a static html side (for example using gatsby js)
During the build process the error   `WebpackError: ReferenceError: window is not defined` shows up. I'm using  "material-table": "^1.62.0",

It was working on an older version of material-table: https://codesandbox.io/s/yj6jnmqpw9?file=/package.json
but seems to be broken by something introduced in the meantime.  There are some leads on the gatsby side, but not enough to fix the error: https://www.gatsbyjs.org/docs/debugging-html-builds/#how-to-check-if-code-classlanguage-textwindowcode-is-defined

Any help is greatly appreciated

hendra-go:
Yeah, i confirmed this bug.

The problem is jsPDF. Related to [Angular universal does not work with jsPDF: ReferenceError: window is not defined](https://github.com/MrRio/jsPDF/issues/2753)

It is not material-table fault, but probably not published new version until above issue fixed ?

zackkrida:
Yeah this is pretty critical—means any use of Material Table in a project with server-side rendering (SSR) could break, especially considering this was released on a minor version bump (so folks deployment processes may automatically bump to this version).

oze4:
Is this still an issue with `v1.63.0`? I know that version fixed a different jsPDF issue.

floAr:
@oze4 I created a quick fix for this and confirmed it is working for the gatsby stack

floAr:
> 
> 
> Is this still an issue with `v1.63.0`? I know that version fixed a different jsPDF issue.

I think the fix was about putting jsPDF into normal depenency instead of dev. I tried it yesterday and the fix was still present. It breaks when importing the jspdf package during a static build process (because it uses 'window' which is undefined without a browser)

hendra-go:
Thanks @floAr , i also confirmed this PR works with Gatsby.

But for me, i will stick to `material-table 1.62.0` because that version had all i need for my current project.

I will upgrade only if my project needs PDF Export function or material-table becomes more modular or at least has an option to choose which export function suites your need.

Probably instead:
`exportButton: true`

We can use something like:
```
exportButton: {
    exportCsv: true,
    exportPdf: true
}
```

willwill96:
Experiencing the same thing with SSR - just going to revert to 1.62.0 until this is resolved

floAr:
[...]

PR #2114, opened 2020-07-01
guarding jsPDF to only import / run when window is present (aka not d…

floAr:
…uring ssr build)

This guards the jsPDF using in a check if we have a present window (aka running in a browser) this will allow to build server side (ignoring the pdf export here) and then rehydrate it back in when the user runs the webpage

## Related Issue

#2112

## Description

Quick fix until jsPDF fixes theeir own issues (https://github.com/MrRio/jsPDF/issues/2753)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,9 @@
     "dist",
     "types"
   ],
+  "publishConfig": {
+    "registry": "https://npm.pkg.github.com/"
+  },
   "babel": {
     "extends": "./configs/.babelrc"
   },
@@ -82,8 +85,6 @@
     "debounce": "^1.2.0",
     "fast-deep-equal": "2.0.1",
     "filefy": "0.1.10",
-    "jspdf": "1.5.3",
-    "jspdf-autotable": "3.5.3",
     "prop-types": "^15.6.2",
     "react-beautiful-dnd": "^13.0.0",
     "react-double-scrollbar": "0.0.15"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "material-table",
-  "version": "1.64.0",
+  "version": "1.65.0",
   "description": "Datatable for React based on https://material-ui.com/api/table/ with additional features",
   "main": "dist/index.js",
   "types": "types/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -85,6 +85,8 @@
     "debounce": "^1.2.0",
     "fast-deep-equal": "2.0.1",
     "filefy": "0.1.10",
+    "jspdf": "1.5.3",
+    "jspdf-autotable": "3.5.3",
     "prop-types": "^15.6.2",
     "react-beautiful-dnd": "^13.0.0",
     "react-double-scrollbar": "0.0.15"
--- a/package.json
+++ b/package.json
@@ -8,9 +8,6 @@
     "dist",
     "types"
   ],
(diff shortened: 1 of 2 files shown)


floAr:
This should fix #2112

src/components/m-table-toolbar.js
```diff
@@ -1,22 +1,22 @@
...
+import Tooltip from "@material-ui/core/Tooltip";
+import Typography from "@material-ui/core/Typography";
+import withStyles from "@material-ui/core/styles/withStyles";
+import { lighten } from "@material-ui/core/styles/colorManipulator";
+import classNames from "classnames";
+import { CsvBuilder } from "filefy";
+import PropTypes, { oneOf } from "prop-types";
+import "jspdf-autotable";
+import * as React from "react";
+const jsPDF = typeof window !== "undefined" ? require("jspdf") : null;
```
ames89, line 19: pretty good

mbrn:
Please resolve conflicts

floAr:
@mbrn & @ames89: All done

package.json
```diff
@@ -82,8 +85,6 @@
     "debounce": "^1.2.0",
     "fast-deep-equal": "2.0.1",
     "filefy": "0.1.10",
-    "jspdf": "1.5.3",
-    "jspdf-autotable": "3.5.3",
```
gamb, line 86: Do we need to remove these? I just ran into an issue with yarn (berry):
> Module not found: material-table tried to access jsPDF, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

gamb: Ah my yarn issue turned out to be due to the dynamic `require()` stuff, so re-adding these dependencies wasn't sufficient anyway.

floAr: I think you are right, while merging something went amiss. The idea is to have those depencies isntalled, but only require them when running withing a browser.

floAr: resolved in
f6b13cb

gamb: Nice one @floAr :)

package.json
```diff
@@ -8,6 +8,9 @@
     "dist",
     "types"
   ],
+  "publishConfig": {
```
mbrn, line 11: Why did you add this?

src/components/m-table-toolbar.js
```diff
@@ -56,15 +58,10 @@ export class MTableToolbar extends React.Component {
   defaultExportCsv = () => {
     const [columns, data] = this.getTableData();
 
-    let fileName = this.props.title || "data";
```
mbrn, line 59: Please rebase your branch. It seems that you removed some new developments.

package.json
floAr: seems like this was an auto add by something in vscode. removed it in 9804e9d

src/components/m-table-toolbar.js
floAr: done

src/components/m-table-toolbar.js
theeranan: thank you

floAr:
@mbrn I am getting some errors with the prettier step:


```
remote: Permission to mbrn/material-table.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/mbrn/material-table/': The requested URL returned error: 403
```

Is this something on my side to fix?

Issue #2077, opened 2020-06-21
Column with arrays of value display as a string


jetro4u:
I have a column with arrays of values in the field. I am using remote data. For example, let's say the values comes in as:
value1, value2, value3. In the column where it appears, I have it as: value1value2value3.

As you can see there is no space or even a comma to separate the values. Is this a bug or is this how it's supposed to work?

Domino987:
Can you share your column definitions and how your data looks like?

PR #2126, opened 2020-07-03
Fix arrays of value display as a string issue #2077

ramazancetinn:
## Related Issue

Relate the Github issue with this PR using `#2077`

## Description

I tried to fix the related issue #2077  

m-table-cell.js

\*

### DIFF
--- a/src/components/m-table-cell.js
+++ b/src/components/m-table-cell.js
@@ -66,6 +66,11 @@ export default class MTableCell extends React.Component {
     } else if (typeof this.props.value === "boolean") {
       // To avoid forwardref boolean children.
       return this.props.value.toString();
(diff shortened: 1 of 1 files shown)


src/components/m-table-cell.js
```diff
@@ -66,6 +66,11 @@ export default class MTableCell extends React.Component {
     } else if (typeof this.props.value === "boolean") {
       // To avoid forwardref boolean children.
       return this.props.value.toString();
+    } else if (Array.isArray(this.props.value)) {
```
mbrn, line 69: What happens if it has different type of elements than string?

mbrn:
Hi @ramazancetinn 

Thanks for your effort. But we let developers to use custom rendering feature on this kind of data.

PR #2144, opened 2020-07-08
Error state for select

### LINKED ISSUE #2142: React does not recognize the `errorState` prop on a DOM element.
After upgrading material-table version 1.54.0 to 1.64.0 it's showing this error: 
` **React does not recognize the `errorState` prop on a DOM element.**`

[
Screenshot from 2020-07-08 12-05-54
](url)

Domino987:
## Description

This adds the error state to the select edit field

## Related PRs

#1998

Also fixes #2142

### DIFF
--- a/src/components/m-table-cell.js
+++ b/src/components/m-table-cell.js
@@ -140,7 +140,7 @@ export default class MTableCell extends React.Component {
   };
 
   render() {
-    const { icons, columnDef, rowData, ...cellProps } = this.props;
+    const { icons, columnDef, rowData, errorState, ...cellProps } = this.props;
     const cellAlignment =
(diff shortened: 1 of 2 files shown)


morwalz:
Hi Please merge the PR soon

darewreck54:
https://codesandbox.io/s/material-table-starter-template-ik4v9?file=/src/index.js The problem still there.  

image


I can still save even when i return a message which is another indication of an error.  This should block the ability to save the row like how it is when you return false.

image

when you return false it greys out the save button

Domino987:
@darewreck54 I missed one line. This will now be fixed with the mentioned PR. sorry for the delay. #2174

PR #2151, opened 2020-07-09
Fix onChange when numeric value

francescovenica:
## Description

Passing `valueAsNumber` instead of `value` when the field is numeric


joaocarlospf:
After this change, the numeric type field is not letting type '.' as you write a decimal number. The only way it accepts decimals is by pasting the final number in the field. I think that instead the value should be converted when the final value will be saved.

francescovenica:
mmmm, let me check, in my app I'm able to type `.` without problem.

PR #2165, opened 2020-07-13
A little refator in material-table.getProps

theqp:
I will explain in the commit comments

### DIFF
--- a/src/material-table.js
+++ b/src/material-table.js
@@ -182,42 +182,23 @@ export default class MaterialTable extends React.Component {
 
     calculatedProps.actions = [...(calculatedProps.actions || [])];
 
-    if (calculatedProps.options.selection)
-      calculatedProps.actions = calculatedProps.actions
-        .filter((a) => a)
-        .map((action) => {
-          if (
-            action.position === "auto" ||
-            action.isFreeAction === false ||
-            (action.position === undefined && action.isFreeAction === undefined)
-          )
-            if (typeof action === "function")
-              return { action: action, position: "toolbarOnSelect" };
-            else return { ...action, position: "toolbarOnSelect" };
-          else if (action.isFreeAction)
-            if (typeof action === "function")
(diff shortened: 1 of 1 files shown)


src/material-table.js
```diff
@@ -182,42 +182,23 @@ export default class MaterialTable extends React.Component {
 
     calculatedProps.actions = [...(calculatedProps.actions || [])];
 
-    if (calculatedProps.options.selection)
```
theqp, line 185: This if have been merged because only two lines differ
```
              return { action: action, position: "toolbarOnSelect" };
            else return { ...action, position: "toolbarOnSelect" };
```
```
              return { action: action, position: "row" };
            else return { ...action, position: "row" };
```

src/material-table.js
```diff
@@ -182,42 +182,23 @@ export default class MaterialTable extends React.Component {
...
-      calculatedProps.actions = calculatedProps.actions
-        .filter((a) => a)
```
theqp, line 187: removed because `a` is never falsy, type of `a` is `Action<RowData> | ((rowData: RowData) => Action<RowData>)`

src/material-table.js
```diff
@@ -182,42 +182,23 @@ export default class MaterialTable extends React.Component {
...
-          if (
-            action.position === "auto" ||
-            action.isFreeAction === false ||
-            (action.position === undefined && action.isFreeAction === undefined)
-          )
-            if (typeof action === "function")
-              return { action: action, position: "toolbarOnSelect" };
-            else return { ...action, position: "toolbarOnSelect" };
-          else if (action.isFreeAction)
-            if (typeof action === "function")
```
theqp, line 198: removed because never true
when action is function it always stops at the first if
```
(action.position === undefined && action.isFreeAction === undefined)
```

mbrn:
Hi @theqp 

Thanks for your effort. But we will start a big refactoring progress. So i won't merge this pr for now.

PR #2174, opened 2020-07-14
Fix for validate with string

Domino987:
## Related Issue

#2144 

## Description

Fixes the validate disabling of the save button for strings correctly.


dbryand:
Thanks for the contributions, @Domino987. Quick question: does validate work for `onAdd` or just `onEdit`? Thanks!

Domino987:
This will work on both.

PR #2176, opened 2020-07-15
fix React.node issue

developerdu:
## Related Issue

Fixes #2171
Fixes #2119

## Description

Should be React.ReactNode and not React.Node

## Related PRs

https://github.com/mbrn/material-table/pull/2119

## Impacted Areas in Application

\* import error

## Additional Notes

n/a

PR #2184, opened 2020-07-16
Fix ReactNode type

### LINKED ISSUE #2183: Type definition has missing type
`React.Node` is not a valid type

https://github.com/mbrn/material-table/blob/15ba31de7ebdeecc563256070ce371b477745faf/types/index.d.ts#L171

v1.65.0 doesn't work with some typescript configs as a result. (eg: CRA)

avin-kavish:
## Related Issue

Closes #2183


mbrn:
Thanks for your effort. Same with #2176

Issue #2186, opened 2020-07-16
Adding Custom Action Component  - The proper way


benedery:
Hey,
I've wanted to add a custom component instead of the button you provide to add a row.

first i was only changing the icon property in the 
```              {
                       icon: () => <AddRowButton />,
                       tooltip: "Add Row",
                       isFreeAction: true,
                       onClick: (event, rowData) => handleDeleteRow(rowData),
                     },
```

BUT it causes a warning of validating DOM with button inside another button.

So I was able to resolve this warning by adjusting the components section
```
                  components={{
                    Action: (props) => {
                      if (props?.action?.tooltip === "Add Row") {
                        return <AddRowButton  />;
                      } else return <MTableAction {...props} />;
                    },
                  }}
```

BUT, Basing my solution on a tooltip looks strange,  
and maybe cause problems in production in the future.  
I want to know if there is a better/safer/cleaner way,  
Thanks.

mbrn:
Hi @benedery 

Please take a look at second example at:

https://material-table.com/#/docs/features/component-overriding

atinyakov:
It seems that @benedery asked about customizing "Add button" which is located in a **header**, not in a row. 
I have the same issue, and it seems that the reply above does not address this. @mbrn please assist.

benedery:
@atinyakov iv'e already make a solution in https://github.com/mbrn/material-table/pull/2206, but they decided to declined, you can use it local if you want.

PR #2206, opened 2020-07-19
Add a specific action custom component

benedery:
## Related Issue

https://github.com/mbrn/material-table/issues/2186

## Description

Added support to add custom action component.

When adding actions Array,  
now each action support isCustom, and Component Properties.  
when isCustom enabled and there is Component supplied.  
it will be assign instead of the normal Button wrapper supplied.
```
                  actions={[
                    {
                      icon: ()=> </>,
                      tooltip: "Delete Row",
                      onClick: (event, rowData) => console.log(rowData),
                      isCustom: true,
                      component: <CustomButton />,
                      isFreeAction:false
                   }]}
```
\*

## Additional Notes

It will be displayed with no errors about Warning: validateDOMNesting or FowardRef.

### DIFF
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -4,6 +4,7 @@ import PropTypes from "prop-types";
 import Icon from "@material-ui/core/Icon";
 import IconButton from "@material-ui/core/IconButton";
 import Tooltip from "@material-ui/core/Tooltip";
+import { MyComponent } from "./m-table-custom-action-component";
 /* eslint-enable no-unused-vars */
 
 class MTableAction extends React.Component {
@@ -47,21 +48,20 @@ class MTableAction extends React.Component {
       );
 
     const button =
-     (action.isCustom || this.props.isCustom && action.component || this.props.component) ?
-         action.component|this.props.component
-         :
-            (
-              <IconButton
-                  size={this.props.size}
-                  color="inherit"
-                  disabled={disabled}
-                  onClick={handleOnClick}
-              >
-                {icon}
-              </IconButton>
-          );
-
-
+      action.isCustom ||
+      (this.props.isCustom && action.component) ||
+      this.props.component ? (
+        <MyComponent>{action.component}</MyComponent>
+      ) : (
+        <IconButton
+          size={this.props.size}
+          color="inherit"
+          disabled={disabled}
+          onClick={handleOnClick}
+        >
+          {icon}
+        </IconButton>
+      );
 
     if (action.tooltip) {
       // fix for issue #1049
--- a/src/components/m-table-action.js
+++ b/src/components/m-table-action.js
@@ -46,16 +46,22 @@ class MTableAction extends React.Component {
         <action.icon />
       );
 
-    const button = (
-      <IconButton
-        size={this.props.size}
-        color="inherit"
-        disabled={disabled}
-        onClick={handleOnClick}
-      >
-        {icon}
-      </IconButton>
-    );
+    const button =
+     (action.isCustom || this.props.isCustom && action.component || this.props.component) ?
+         action.component|this.props.component
+         :
+            (
+              <IconButton
+                  size={this.props.size}
+                  color="inherit"
+                  disabled={disabled}
+                  onClick={handleOnClick}
+              >
+                {icon}
+              </IconButton>
+          );
+
+
 
     if (action.tooltip) {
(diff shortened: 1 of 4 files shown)


mbrn:
MT already has component overriding. You can use it: https://material-table.com/#/docs/features/component-overriding

benedery:
> MT already has component overriding. You can use it: https://material-table.com/#/docs/features/component-overriding

but it for all the actions, if a user want to keep the table action component and only one item to be custom?( a real use case)

mbrn:
Can you try this:

```javascript
<MaterialTable
      title="Action Overriding Preview"
      columns={[
        { title: 'Surname', field: 'surname' },
        {
          title: 'Birth Place',
          field: 'birthCity',
        },
      ]}
      data={[
        { name: 'Mehmet', surname: 'Baran', birthYear: 1987, birthCity: 63 },
        { name: 'Zerya Betül', surname: 'Baran', birthYear: 2017, birthCity: 34 },
      ]}
      actions={[
        {
          icon: 'save',
          tooltip: 'Save User',
          onClick: (event, rowData) => alert("You saved " + rowData.name)
        },
        {
          icon: 'refresh',
          tooltip: 'Update User',
          onClick: (event, rowData) => alert("You updated " + rowData.name)
        }
      ]}
      components={{
        Action: props => {
          if(props.action.icon === 'save') {
            return (
              <Button
                onClick={(event) => props.action.onClick(event, props.data)}
                color="primary"
                variant="contained"
                style={{textTransform: 'none'}}
                size="small"
              >
                My Button
              </Button>
            );
          }
          return "default button";
        },
      }}
    />
```

You can reach default action with
`import { MTableAction } from "material-table"`

fpw23:
@benedery Another option that might be cleaner is to override the Action Component and make it smart so that depending on the props it can render different components.  This is what I have done so that I can render buttons in rows but icons in the header:

```
class MTableAction extends React.Component {
  render () {
    let action = this.props.action

    if (typeof action === 'function') {
      action = action(this.props.data)
      if (!action) {
        return null
      }
    }

    if (action.action) {
      action = action.action(this.props.data)
      if (!action) {
        return null
      }
    }

    if (action.hidden) {
      return null
    }

    const disabled = action.disabled || this.props.disabled

    const handleOnClick = (event) => {
      if (action.onClick) {
        action.onClick(event, this.props.data)
        event.stopPropagation()
      }
    }

    const icon =
      typeof action.icon === 'string' ? (
        <Icon {...action.iconProps}>{action.icon}</Icon>
      ) : typeof action.icon === 'function' ? (
        action.icon({ ...action.iconProps, disabled: disabled })
      ) : (
        <action.icon />
      )
```
Here is the part I change

 ```
   const button = action.position === 'row'
      ? <Button
        variant='outlined'
        disabled={disabled}
        onClick={handleOnClick}
      >
        {action.title}
      </Button>
      : <IconButton
        size={this.props.size}
        color="inherit"
        disabled={disabled}
        onClick={handleOnClick}
      >
        {icon}
      </IconButton>

```
   ```
 if (action.tooltip) {
      // fix for issue #1049
      // https://github.com/mbrn/material-table/issues/1049
      return disabled ? (
        <Tooltip title={action.tooltip}>
          <span>{button}</span>
        </Tooltip>
      ) : (
        <Tooltip title={action.tooltip}>{button}</Tooltip>
      )
    } else {
      return button
    }
  }
}

MTableAction.defaultProps = {
  action: {},
  data: {}
}

MTableAction.propTypes = {
  action: PropTypes.oneOfType([PropTypes.func, PropTypes.object]).isRequired,
  data: PropTypes.oneOfType([
    PropTypes.object,
    PropTypes.arrayOf(PropTypes.object)
  ]),
  disabled: PropTypes.bool,
  size: PropTypes.string
}

export default MTableAction
```

I am able to add a title prop to the action item in the list so that I can set the text for the button
```
actions={[
              {
                title: 'View',
                position: 'row',
                tooltip: 'View Details',
                onClick: (event, rowData) => { this.onUpdateClick(rowData._key, event.refreshTableData) }
              }
]}
```

I am switching the result based on the position prop but if you need even more customization there is nothing stopping you from adding a type prop with a string then in the main code switch case on the type and render any number of different results.

benedery:
@fpw23 thank you for your suggestion, my pr solved me and my team the problem but i wanted to help others.

PR #2210, opened 2020-07-20
fix: Fixed EditComponentProps

TriPSs:
EditComponentProps had the wrong error defined in the interface.


mbrn:
@Domino987 

Can you check change please?

kkirsche:
With this change, where do you retrieve the error message now?

PR #2214, opened 2020-07-21
set currency as number instead of text

francescovenica:
## Description

Set currency as numeric field


Domino987:
Hi thanks for the contribution. Can you validate what happens if the field gets cleared? Will there be a 0 or an empty field ?

francescovenica:
Hello I kept the same behaviour of the numeric textfield, if the value is undefined it use an empty string

francescovenica:
ok, I understood what do you mean, with no value it gives NaN, I'll fix it!

Domino987:
Looks better. I'll test it later. Can you remove the demo changes though?

francescovenica:
sure!

Issue #2168, opened 2020-07-13
A way to not include jspdf


niicojs:
Jspdf is 30% of material-table bundle size.  
PDF export is a nice feature but it would be nice to avoid including this lib when export is not needed. It would greatly benefits bundle size.

I don't know if getting it tree-shaking compatible is possible or 2 versions are needed but it would be awesome 😁

niicojs:
I've just noticed bundle size almost double recently with 1.63.

https://bundlephobia.com/result?p=material-table@1.64.0

johnmagbag1995:
Having same problem #2164

niicojs:
Correct you opened the issue between the time I searched for previous issues and the time I opened it 😅

Issue #2164, opened 2020-07-13
Large bundle size (jspdf)


johnmagbag1995:
**Jspdf** has 290kb of bundle size even though I am not using it. Is it possible to remove it from bundle?

table options: 
    pageSize: 10,
    actionsColumnIndex: -1,
    sorting: true,
    draggable: false,
    search: false,
    showTitle: false,
    toolbar: false,
    filtering: false,

![image](https://user-images.githubusercontent.com/61783659/87300145-eee7c680-c53f-11ea-831e-3ed778a07630.png)

avin-kavish:
Btw, how do I enable the PDF export option?

shovelend:
I can't build the app with Next.js ever since jspdf has been added as a dep. Looks like jspdf can only run client-side not server side!
We don't need the Export as PDF feature (never asked for it), at least there should be a way to exclude it if you guys aren't thinking about removing it.

niicojs:
A way to fix all that would be to move `defaultExportPdf ` to a new module, which shouldn't be included by default.
The lib won't have a default export function, but it could be included using `exportPdf={defaultExportPdf}`.
Done that way, tree shaking should work.

akhilmhdh:
Hey guys so is this getting fixed ...or is there a way to avoid the bundle

linchen1987:
the tree shaking doesn't work in my code
```
import MaterialTable from 'material-table'
```
how to write code to implement tree shaking

seromenho:
Maybe export to pdf should live in user land and a sample on how to could be provided instead or this extracted to another package and that one having the jspdf dependency.
I've seen there's already a `exportCsv` prop you can in as a custom function, what do you think if jspdf is completely removed and a sample added on how to achieve such export? (such sample or package is mostly done with the [defaultExport](https://github.com/mbrn/material-table/blob/166f19746ff1811e3d37e22d9337df9caf059ec7/src/components/m-table-toolbar.js#L77) already)
Same for export as csv.

niicojs:
That's what I did in my PR. Moving export function to a different file that you need to opt in.
So it gets tree shake if not used.

seromenho:
Well it's similar then, not the same. I've entirely removed default exporters. Not sure though if it will be considered since it seems a refactor is going on.

seromenho:
[...]

PR #2221, opened 2020-07-22
Allow tree shaking to remove export dependencies if not used

niicojs:
## Related Issue

#2168 and #2164

## Description

Move default export function into a different file.  
This will reduce file size a lot (remove jspdf and csvbuilder).  
It will also remove default export function which will have to be defined manually

- Remove default export functions
- Examples for exports will have to be updated

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -2,8 +2,10 @@ import { Grid, MuiThemeProvider, Button } from "@material-ui/core";
 import { createMuiTheme } from "@material-ui/core/styles";
 import React, { Component } from "react";
 import ReactDOM from "react-dom";
-import MaterialTable from "../src";
 import Typography from "@material-ui/core/Typography";
+
+import MaterialTable from "../src";
+import { defaultExportCsv, defaultExportPdf } from "../src/exports";
 
 let direction = "ltr";
 // direction = 'rtl';
@@ -494,6 +496,9 @@ class App extends Component {
                     console.log("selected Filters : ", appliedFilter);
                   }}
                   options={{
+                    exportButton: true,
+                    exportCsv: defaultExportCsv,
+                    exportPdf: defaultExportPdf,
                     headerSelectionProps: {
                       color: "primary",
                     },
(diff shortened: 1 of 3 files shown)


mbrn:
Hi @niicojs 

Thanks for your effort. But we are working on refactoring and tests. So i can't approve this PR. Thanks for your understanding.

avin-kavish:
@mbrn so, what is a viable solution for us while you do the refactoring? Does that refactoring involve being able to drop jsPDF when not needed?

johnmagbag1995:
@avin-kavish I just added jspdf into webpack's externals for temporary while waiting for the new version.

avin-kavish:
@johnmagbag1995 thanks

Forsaiken:
How can I remove jspdf from bundle? this is huge and totally unnecessary

avin-kavish:
@johnmagbag1995 is that possible with CRA? AFAIK, it's not possible to customize the webpack config with it.

avin-kavish:
@niicojs If I fork this, will it definitely work?

avin-kavish:
@mbrn I don't get it, this seems like a perfectly reasonable solution, I'm sure this could be merged and refactored late if needed.

baocancode:
Hi, any plan to make jspdf optional in/after the refactoring?

diegosps:
In production build, I have:
jspdf 290 kb
react-beautiful-dnd 85 kb
data-fns: 49kb
jspdf-autotable: 31 kb

There are more, but with lower sizes.

So it's at least 455 kb added on top the 136 kbs from core for features that I'm not sure all of its user base have a need.
This is taxing on slow connections.

I agree with avin-kavish, even though it's a breaking change, at least for me, it is better than having 320 kb of unused code.

I hope these changes make its way in the refactor.

@avin-kavish Did you find a suitable way for CRA without eject? I found [this](https://auth0.com/blog/how-to-configure-create-react-app/) and [this](https://github.com/harrysolovay/rescripts) as possible solutions.

Issue #2212, opened 2020-07-20
Persist filters after update data.


kecoco16:
Hi, I am using material-table with real time data(subscriptions), I have the following scenario and I want to know if there is any way to solve it:

I want to persist the filters after the data is updated: currently when new data is added by the subscription all the filters are removed

![demo](https://user-images.githubusercontent.com/25941813/87995417-adf63180-caac-11ea-8462-daa749e6879d.gif)

[CodeSandbox Demo](https://codesandbox.io/s/silly-hamilton-m0zyf?file=/src/App.js)

PR #2222, opened 2020-07-22
feat: persist tableData

kecoco16:
## Related Issue
#2212 

## Description

Adding the ability to persist filters when properties change very useful when used with real-time data.


mbrn:
Hi @kecoco16 

Thanks for effort. But we have good plans for tableData field on both rowData and columnDef objects. So i can't merge this pr. 

Thanks for your understanding.

PR #2008, opened 2020-06-08
Approve edit/add on enter added

vijaypushkin:
## Related Issue
#1364 
#1162 

## Description
This PR will add the feature to save edit/add upon pressing enter key

## Related PRs
#1691 

## Impacted Areas in Application
List general components of the application that this PR will affect:

* Edit Row
* Add Row

Issue #1364, opened 2019-11-20
How to submit onRowAdd form with Enter key?


rommelmamedov:
Hi there! Is there any possible way to catch enter key press to submit the onRowAdd form or confirm onRowDelete?

MohitKS5:
@vijaypushkin  Now that the PR is merged, how to prevent this ? I don't want to bind enter keys to save, enter is being used for some other purponse.

Issue #1162, opened 2019-10-05
How to support enter and escape key


Hadsch:
For a better Usability I would like to support enter and escape key.

Therefore I can use a custom edit componet:
   
```
  columns={[
                        { title: "Just a title", field: "name", editComponent: props => (
                          <Input
                            type="text"
                            autoFocus={true}
                            value={props.value}
                            onChange={e => props.onChange(e.target.value)}
                            onKeyPress={this.handleKeyPress}
                          />
                        ) },
```

The problem is, that I don't know how to call the onEditingApproved method inside the handleKeyPress method.

```
handleKeyPress = (event) => {
    if(event.key === 'Enter'){
      console.log('enter press here! ')
      // how to fire "onEditingApproved = (mode, newData, oldData)" event from here?
    }
  }
```

Generally I think this would be a great feature if this would be a default behavior! 
If you want I can submit a PR.

As far as I can see, the EditComponent itself should already listen to the enter and escape key

flyingangel:
There is no onEditingApproved method. 
But I'd like to have feature to trigger update/add as well.

jayarjo:
Yeah, that'd be nice.

jiangyh123456:
```
  components={{
            EditField: props => {
              console.log(props);
              return (
                <TextField
                  onChange={e => {
                    props.onChange(e.target.value);
                    if (e.target.value) {
                      if (tagNameField.invalid) {
                        setTagNameField({
                          invalid: false
                        });
                      }
                    } else {
                      if (!tagNameField.invalid) {
                        setTagNameField({
                          invalid: true
                        });
                      }
                    }
                  }}
                  onKeyUp={event => {
                    if (event.key === "Enter") {
                      tableRef.current.onEditingApproved(
                        props.rowData.id ? "update" : "add",
                        props.rowData
                      );
                      console.log(tableRef.current);
                    }
                  }}
                  value={props.value || ""}
                  autoFocus
                  error={tagNameField.invalid}
                  helperText="tag name"
                />
              );
            }
          }}
```

jiangyh123456:
call tableRef.current.onEditingApproved()  and pass "update" or "add" and rowData
seems to work. I did not pass the third param "old data"

Domino987:
Escape is now supported #1691

PR #2227, opened 2020-07-23
Fixed  Approve onEnter overriding textarea behavior

vijaypushkin:
## Related Issue

#1364
#1162

## Description

Fixed a bug from PR #2008 
* Approve save onEnter will be skipped for text-area a.k.a Material UI TextField mutliline
* Added shift + enter for approve save added for text-area

## Related PRs

#2008 

* Edit Row
* Add Row


DeepikaVelagandula:
Save on enter should also be prevented for lookup having multiple values in it

PR #2242, opened 2020-07-26
Feature 1471 editable cell

mbrn:


lfsando:
#2244 I believe this issue has something to do with this feature

PR #2255, opened 2020-07-30
Missing hiddenByColumnsButton in Column type 

Shaquu:
Column type is missing hiddenByColumnsButton property.

## Description

Without that field you cannot set column to be hidden by default and also available to be shown with "Add or remove columns" window.

- "Add or remove columns" window


Shaquu:
@mbrn Please assist as it stops me.

PR #2289, opened 2020-08-06
Add Editing component props to type

kkirsche:
## Related Issue

N/A

## Description

Add support for editing functions to allow users of typescript to override editing-related components such as MTableEditRow. This also adds the mode's as a portion of the props, as it's used by the editing functions

Typescript type definition

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -45,10 +45,13 @@ export interface MaterialTableProps<RowData extends object> {
   options?: Options<RowData>;
   parentChildData?: (row: RowData, rows: RowData[]) => RowData | undefined;
   localization?: Localization;
+  mode?: EditMode;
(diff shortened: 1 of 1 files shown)


kkirsche:
Prettier failed to checkout the repo it seems, not an actual failure

types/index.d.ts
```diff
@@ -45,10 +45,13 @@ export interface MaterialTableProps<RowData extends object> {
   options?: Options<RowData>;
   parentChildData?: (row: RowData, rows: RowData[]) => RowData | undefined;
   localization?: Localization;
+  mode?: EditMode;
```
mbrn, line 48: mode is not props.

mbrn:
Thanks for your effort. But table does not accept a prop 'mode'

Issue #2178, opened 2020-07-15
Error installing v 1.64.0 and up


andyvanhelsing:
To reproduce:
npm install material-table@1.64.0

npm ERR! code ENOENT
npm ERR! syscall spawn git
npm ERR! path git
npm ERR! errno ENOENT
npm ERR! enoent Error while executing:
npm ERR! enoent undefined ls-remote -h -t ssh://git@github.com/eligrey/FileSaver.js.git
npm ERR! enoent
npm ERR! enoent
npm ERR! enoent spawn git ENOENT
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent

Not seen when installing v 1.63.0 and earlier

stevestovold:
Also getting this issue

batsauto:
This is related to Issue #2109 

Which was caused by this PR https://github.com/mbrn/material-table/pull/1970

usahai:
Please resolve! Trying to update from 1.57.2 to get some new features but not able to upgrade!

vmmelo:
Any workaround?

batsauto:
in my package.json i made this edit
`"material-table": "1.62.0",`

**notice the removal of the ^** carrot. 

This is the last major version before using jsPDF that i know of.

ZakKa89:
Thank you @batsauto , we were manually changing the package lock and that was becoming tedious

chmelevskij:
When using meteor js and upgrading to `1.69.0` had following issue when running tests:
```
TypeError: Cannot assign to read only property 'then' of object '#<t>'
```
Downgrading back to:
```
"material-table": "1.62.0",
```
Solved the issue, leaving comment here for anyone else running into this cryptic issue with meteor

DRsabadash:
For anyone stumbling across this, a workaround to be able to install v 1.64.0 or greater is to ensure you have git installed in your environment. 

if you're installing/running un-dockerized, just make sure git is in your PATH and accessible from the command line.
if you're installing/running dockerized, add the line RUN apk add --no-cache git to your Dockerfile before any npm commands.

Issue #2137, opened 2020-07-07
NPM ISSUE


ashindiano:
npm ERR! code ENOENT
npm ERR! syscall spawn git
npm ERR! path git
npm ERR! errno -2
npm ERR! enoent Error while executing:
npm ERR! enoent undefined ls-remote -h -t ssh://git@github.com/eligrey/FileSaver.js.git
npm ERR! enoent 
npm ERR! enoent 
npm ERR! enoent spawn git ENOENT
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 

- #### Please try to be as detailed as possible
npm install gives ssh error when tried to install  jspdf 1.5.3

**Describe the bug**
Enterprise Networks do not allow ssh. But material table uses jspdf 1.5.3 which uses github:eligrey/FileSaver.js#1.3.8
enterprise networks does not allow ssh to FileSaver

**To Reproduce**

npm install

**Expected behavior**

ashindiano:
It was because ssh client was not installed in Docker. Wrong assessment on the issue. So closing it

black-puppydog:
I just stumbled over this issue as well and frankly I don't see why an ssh-agent should be required to install publicly available packages. Is this something that could be solved by specifying the requirements differently?

Issue #2109, opened 2020-06-29
Docker build fails on `npm install: Error while executing: undefined ls-remote -h -t ssh://git@github.com/eligrey/FileSaver.js.git


mikeestradathd:
I am trying to build a Docker image but I get the following error:

```
Error while executing:
npm ERR! enoent undefined ls-remote -h -t ssh://git@github.com/eligrey/FileSaver.js.git
```

Simply running `docker build .`
When I take out material-table, the build passes.

gv24z:
I have this problem too:
`> npm install material-table
npm ERR! code ENOENT
npm ERR! syscall spawn git
npm ERR! path git
npm ERR! errno ENOENT
npm ERR! enoent Error while executing:
npm ERR! enoent undefined ls-remote -h -t ssh://git@github.com/eligrey/FileSaver.js.git
npm ERR! enoent
npm ERR! enoent
npm ERR! enoent spawn git ENOENT
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\user\AppData\Roaming\npm-cache\_logs\2020-06-30T08_32_59_747Z-debug.log`

scscgit:
This seems to be caused by a dependency on Git being installed on the machine, so you can add this to your Dockerfile:

> \# Git needed by material-table
> RUN apk add git

Hopefully this can be fixed in the future, so there's no such dependency required by default.

Here's one additional error message:
npm ERR! missing: file-saver@github:eligrey/FileSaver.js#1.3.8, required by jspdf@1.5.3

The dependency of material-table on jspdf@1.5.3 can be seen by this dependency graph:
+-- material-table@1.63.1
| +-- @date-io/date-fns@1.3.13 deduped
| +-- @material-ui/pickers@3.2.10 deduped
| +-- classnames@2.2.6
| +-- date-fns@2.14.0 deduped
| +-- debounce@1.2.0
| +-- fast-deep-equal@2.0.1
| +-- filefy@0.1.10
| +-- jspdf@1.5.3

And this library seems to be used here:
https://github.com/mbrn/material-table/blob/22d8fca70c6d45b6035ca8972f8a0fce6ac350c5/src/components/m-table-toolbar.js#L73

mikeestradathd:
@scscgit 
`RUN apk add git` works for an Alpine distribution, what would be the equivalent for `node:12`?

[...]

PR #2310, opened 2020-08-12
bump jspdf to latest version

seromenho:
## Related Issue

#2109 #2137 

## Description

jspdf got a new [2.0.0 release ](https://github.com/MrRio/jsPDF/releases/tag/v2.0.0) 

Im not really sure if it's affected, maybe export to pdf only.

## Additional Notes

Maybe jspdf should be optional for the lib as I've seen in some other issues. If there is some plan for that this will fixe related issues while it's not implemented.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "debounce": "^1.2.0",
     "fast-deep-equal": "2.0.1",
     "filefy": "0.1.10",
-    "jspdf": "1.5.3",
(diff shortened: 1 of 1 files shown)


bigianb:
same as #2311 except this one does not update jspdf-autotable.
Should resolve #2178

mbrn:
Thanks for your effort. But i will merge #2311

PR #2312, opened 2020-08-12
BREAKING CHANGE: remove default exporters

seromenho:
## Related Issue

#2164 #2168

## Description

Removes default csv and pdf exporters. Goal is to reduce bundle size for those who don't use the feature. (Or maybe most people use the feature 🤷🏻‍♂️) 

Everyone using default exporters will be affected. This is a BREAKING CHANGE.

## Additional Notes

- default exporters can become a material-table package eg: "@material-table/pdf-export"
- maybe export button can be removed also and give place to one people can customize


a-tonchev:
@seromenho Please try while refactoring also to reduce possibly the size of material-table. It is still quite big - the package is big like the whole material-ui

mbrn:
Please can you take a look to conflicts?

seromenho:
@mbrn conflicts resolved.

seromenho:
@mbrn like I have mentioned on the Additional Notes, is it an option to create a package with the current default exporters?

PR #2344, opened 2020-08-21
support export all remote data

chunkitmax:
## Related Issue

N/A

## Description

Currently, the default export functions export data loaded on the current page.
Even when `exportAllData: true`, the table exports the rendered remote data only (on the current page).
This patch allows it to get all remote data for export.

## Additional Notes

Not sure if async/await is allow to use :)

### DIFF
--- a/src/components/m-table-toolbar.js
+++ b/src/components/m-table-toolbar.js
@@ -46,7 +46,7 @@ export class MTableToolbar extends React.Component {
         a.tableData.columnOrder > b.tableData.columnOrder ? 1 : -1
       );
     const data = (this.props.exportAllData
-      ? await this.getAllData()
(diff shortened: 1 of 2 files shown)


chunkitmax:
Hi @ollyrowe, thank you for your suggestion. Yes, it should send a single request for all data.

robincsamuel (approved):

ollyrowe:
@Domino987, is there any progress with this?

prevostc:
Nice addition !

robincsamuel:
Won't fix?

Domino987:
Hello @robincsamuel Since mbrn is afk, we moved to material-table/core to support this library further. Please update to the new lib and see if it works for you. Thank you

PR #2346, opened 2020-08-21
Restore editvalues on bulk edit paging

Domino987:
## Description

This PR adds the restoring of the bulkEdited values on paging.
Without this, the default values are shown if the user navigates back to previously edited page, instead of the edited value.

For that, to access the current bulk editValue, getFieldValue is modified to return the last bulk edit data if data is present.
This returned value updates the passed renderData to show these new last edited values.

* TableBody
* Data-Manager


Domino987:
+1

shamseerahammedm:
+1

PR #2368, opened 2020-08-26
Update index.d.ts

jmourell:
Update type definition of customSort to reflect function signature

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -159,7 +159,8 @@ export interface Column<RowData extends object> {
   customSort?: (
     data1: RowData,
     data2: RowData,
-    type: "row" | "group"
+    type: "row" | "group",
+    isDesc: "desc" | undefined
   ) => number;
   defaultFilter?: any;
   defaultGroupOrder?: number;


types/index.d.ts
```diff
@@ -159,7 +159,8 @@ export interface Column<RowData extends object> {
   customSort?: (
     data1: RowData,
     data2: RowData,
-    type: "row" | "group"
+    type: "row" | "group",
+    isDesc: "desc" | undefined
```
mbrn, line 163: Thanks for your effort. But i think it was a mistake. Because we don't need to pass isDesc parameter to customSort function. Because we already check for sort direction before calling.

jmourell:
Sorry I should have explained the use case I'd like to use a custom sort
that ideally takes  advantage of the types defined within your library but
also needs is descending to do correct sorts. That parameter can of course
be optional but I'd like to be there so typescript doesn't fail when I pass
it a function with that parameter defined.

Julian

2020年8月31日(月) 午前1:29 Mehmet Baran <notifications@github.com>:

PR #2396, opened 2020-08-30
Trim search text

### LINKED ISSUE #2394: Search with Leading or Trailing whitespace show "No records to display"
Searching in Material Table search form with leading whitespace or trailing whitespace doesn't display any data. 
For example if my search keyword is `" something "` instead of `"something"` it shows "No records to display".

Even if I use Remote data and change the **TableData** it doesn't show anything. 
I tried trimming the keyword with `trim()` but it doesn't work either.

Domino987:
## Related Issue

Fixes #2394

## Description

This trims the searchtext so that ' A' finds also A.

* Data-Manager


mbrn (approved):
Thanks

PR #2406, opened 2020-09-02
Fix `inputProps` incorrectly nested under `InputProps`

glassblowerscat:
## Description

In `MTableEditField`, some `material-ui` input components were being passed
`inputProps`, which are for passing attributes to the rendered `input` element,
inside `InputProps`, which are for passing props to the wrapped `material-ui`
`Input` component. This resulted in the `aria-label` for these components
(which was all that was being passed in `inputProps`) being incorrectly applied,
throwing a development error and failing to render the `aria-label` in the DOM.

- `MTableEditField`

### DIFF
--- a/src/components/m-table-edit-field.js
+++ b/src/components/m-table-edit-field.js
@@ -123,10 +123,10 @@ class MTableEditField extends React.Component {
           InputProps={{
             style: {
               fontSize: 13,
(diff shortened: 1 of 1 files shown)


mbrn (approved):
Thanks for contribution

PR #2423, opened 2020-09-07
Fixes infinity loop

Domino987:
## Description
This PR fixes an infinity loop generated by the while loop, where the index is increased each turn and will always be above -1. This happens if the table is mounted repeatably in fast succession.

[Link to SO](https://stackoverflow.com/questions/63602354/material-ui-react-table-hangs-on-switching-tabs)

* common values


mbrn (approved):
Thanks for contribution

PR #2428, opened 2020-09-08
Fixed #2427: Optional function for editable is now cheked if editable

### LINKED ISSUE #2427: The `editable` field in the ColumnDef does not work if a function is specified
The editable field in the column definition is supposed to support a function or a string literal but if a function is specified then the function is not called in order to determine if the field should be rendered as editable or not.

**To Reproduce**

The following code snippet can be used to reproduce the problem:

```
interface RowData {
  animal: string;
  count: number;
}

const columns: Column<RowData>[] = [
  { title: "Animal", field: "animal", editable: "never" },
  { title: "Count", field: "count", editable: (_, row) => {
    return false /* This doesn't have any affect */
  }},
];

const data: RowData[] = [
  { animal: "Cow", count: 20 },
  { animal: "Pig", count: 30 },
  { animal: "Duck", count: 50 },
  { animal: "Horse", count: 5 },
  { animal: "Chicken", count: 40 },
];

export const Default = () => (
  <BaseMaterialTable
    columns={columns}
    data={data}
    cellEditable={{
      onCellEditApproved: (newValue, _1, _2, _3) => {
        return new Promise((resolve, reject) => {
          console.log("newValue: " + newValue);
          setTimeout(resolve, 1000);
        });
      },
    }}
  />
);
```

**Expected behavior**

It's expected that the `count` field would not be rendered as editable if the function above always returns `false` and if you alter the function to return `true` then the field should be rendered as editable.  This code shows that the field is rendered as editable no matter what the return value is.

- OS: macOS Catalina - 10.15.5
- Browser Chrome v85
- material-table version: 1.69.0

stgraham2000:
Fixes #2427

## Description

The code was only checking if the editable field was 'never'.  It can also be a function and the result of that function should be used, if specified, to determine the state of the editable.

### DIFF
--- a/src/components/m-table-body-row.js
+++ b/src/components/m-table-body-row.js
@@ -66,7 +66,7 @@ export default class MTableBodyRow extends React.Component {
               }
               rowData={this.props.data}
               cellEditable={
-                columnDef.editable !== "never" && !!this.props.cellEditable
+                ((typeof columnDef.editable !== "function" && columnDef.editable !== "never") || (typeof columnDef.editable === "function" && columnDef.editable(columnDef, this.props.data))) && !!this.props.cellEditable
               }
               onCellEditStarted={this.props.onCellEditStarted}
               scrollWidth={this.props.scrollWidth}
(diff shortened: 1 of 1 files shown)


correaricardo:
@mbrn, this pull request fix a bug with a few lines changes, can you approve it?

gardengeek99:
Unfortunately, I am blocked by this as well.

gardengeek99:
Any ideas when this will be resolved?

Issue #2270, opened 2020-08-04
Save action triggered with invalid data on Enter key pressed


mrjdavidfg:
**Describe the bug**
On Editable feature, if some column has the  `validate` property and you focus on some input and press Enter, the save action is triggered (`onRowAdd`, `onRowUpdate`) even if the row has invalid data.

**To Reproduce**
1. Go to https://codesandbox.io/s/cranky-morse-eo9gf
2. Click on "Add" row icon.
3. Click on some input.
4. Press 'Enter' key.
5. The row was added with empty (invalid) data.

**Expected behavior**
Pressing enter should not trigger save action with invalid data.

jscotty723:
Having this same issue. Is there a way to work around or fix this?

awesomeToan:
Any update on it ?

alpcoker:
I have a [PR proposal](https://github.com/mbrn/material-table/pull/2440) that fixes that bug.  If it is that urgent, you may checkout my [branch ](https://github.com/alpcoker/material-table/tree/hotfix/validate-on-enter) and build your own version with my code. @mrjdavidfg @jscotty723 @awesomeToan 

@mbrn do you have time to review my PR ?

JuddAO19:
Any update on this?

Ashwarrior:
I'm having the same issue. Any update on this??

GustafFig:
I am having the same issue. Some update on this?

Vishal0612:
Any update on this issue pls help

PR #2440, opened 2020-09-10
Fixed Bug : Pressing Enter While Editing Row Bypass Validations

alpcoker:
## Related Issue

#2270 

## Description

Previously when there is validations on any column, we can submit invalid data by pressing enter.  Validate functions are not triggered, new data is directly passing to `onEditingApproved` function.

I move the validation check logic on `renderActions` to a function. I reuse that function on `handleSave` function.

### Example Fail Case By Screenshots

1. Create a validation for a column like
`validate: ({ name }) => name?.trim().length > 3,`

2. As following screenshot `Save` button is disabled
invalid-state

3. But if we press enter invalid data will be submitted
invalid-data-saved

4. After our table has invalid data, we can't delete it also
invalid-data-cannot-be-deleted

m-table-edit-row.js

### DIFF
--- a/src/components/m-table-edit-row.js
+++ b/src/components/m-table-edit-row.js
@@ -160,6 +160,9 @@ export default class MTableEditRow extends React.Component {
   }
 
   handleSave = () => {
+    if (!this.isValid()) {
+      return;
+    }
     const newData = this.state.data;
     delete newData.tableData;
     this.props.onEditingApproved(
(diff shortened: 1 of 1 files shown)


tuliolima:
Someone is working on this PR? I really need this fix.

alpcoker:
In our company, we use our my local version of my branch: https://github.com/alpcoker/material-table/tree/hotfix/validate-on-enter . @tuliolima if it is that urgent you may use that version. 

There were some issues like #2270 , @mbrn or @Domino987 any chance to review my proposal of that tiny but critical bug fix?

tagkiller:
Hi, is it really ok to have closed that ticket ? The bug hasn't disappeared.

Domino987:
Since this project is abandoned, upgrade to @matrrial-table/core

Issue #2433, opened 2020-09-09
Poor performance. Page gets unresponsive


DewangS:
In my functional component, I've a simple table which gets populated via axios call to my NodeJs(Express) based API with minimal data. In my table I don't do any fancy rendering etc. I am getting severe lagging in page rendering. I've debugged the code and can confirm that there are no issues with my APIs as they work perfectly fine with native Material UI table. For now I am switching back to MUI native tables. 

My package details are as per below ...

**"@material-ui/core": "^4.11.0",
"@material-ui/icons": "^4.9.1",
"material-table": "^1.69.0",
    "react": "^16.13.1",
    "react-dom": "^16.13.1",**

Browser : Chrome Version 85.0.4183.83 (Official Build) (64-bit)

Env. : MacOS Mojave 10.14.5

riyaarora65:
I am also getting the same issue from tomorrow.

sharmaadvay0:
Same issue.

JongtaekChoi:
I think it may work to remove the `tableData` attribute of each item of the columns of props at render function.

peacechen:
This was fixed in PR #2689

Unfortunately mbrn is inactive and not merging PR's.  Use the official community fork [@material-table/core](https://github.com/material-table-core/core) instead.

dalbir:
For me rowStyle is the problem, for now removed that. But I think @mbrn should merge the PRs

jhoanborges:
> For me rowStyle is the problem, for now removed that. But I think @mbrn should merge the PRs

This was my solution too. Thank you so much

Issue #2381, opened 2020-08-28
setColumns generates width strings which crash the browser


peragwin:
https://github.com/mbrn/material-table/blob/6b2e168f8bb768b1582998377ed6363abcb0ad71/src/utils/data-manager.js#L87

This is definitely incorrect. Its recursively generating 80+ kb strings in my browser before crashing.

PR #2441, opened 2020-09-10
Fix an error that makes the width value string very long.

JongtaekChoi:
## Related Issue

#2381 
#2433 (not sure)

## Description

`columnDef.tableData.width`may be defined even  if `columnDef.width` is undefined.
Because the calculated `tableData.width` added to each `column` object when the props updated or added.
This function produces a string value with a previously calculated value each time it is called, so the length increases proportionally to the power of the number of columns with has undefined width.
스크린샷 2020-09-11 오전 2 03 40

### DIFF
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -84,7 +84,10 @@ export default class DataManager {
         id: index,
       };
 
-      if (columnDef.width !== undefined) {
+      if (
+        !undefinedWidthColumns.includes(columnDef) &&
+        columnDef.tableData.width !== undefined
+      ) {
         usedWidth.push(columnDef.tableData.width);
       }
 
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -84,7 +84,7 @@ export default class DataManager {
         id: index,
       };
 
-      if (columnDef.tableData.width !== undefined) {
+      if (columnDef.width !== undefined) {
         usedWidth.push(columnDef.tableData.width);
       }
 


cdruck (changes requested):
lines 87 and 88 are inconsistent

JongtaekChoi:
> lines 87 and 88 are inconsistent

How about this way?

joacub (commented):
does not work still the bug

joacub (changes requested):
the problem is this:
**OLD WRONG:**
```js
columnDef.tableData = {
        columnOrder: index,
        filterValue: columnDef.defaultFilter,
        groupOrder: columnDef.defaultGroupOrder,
        groupSort: columnDef.defaultGroupSort || 'asc',
        width: typeof columnDef.width === 'number' ? `${columnDef.width}px` : columnDef.width,
        initialWidth: typeof columnDef.width === 'number' ? `${columnDef.width}px` : columnDef.width,
        additionalWidth: 0,
        id: index,
       **...columnDef.tableData,** <------------------------------------  Mutable data overwrite the width
      };
```

```js
columnDef.tableData = {
        **...columnDef.tableData,** <------------------------------------  Now mutable data is overwrites by the next code
        columnOrder: index,
        filterValue: columnDef.defaultFilter,
        groupOrder: columnDef.defaultGroupOrder,
        groupSort: columnDef.defaultGroupSort || 'asc',
        width: typeof columnDef.width === 'number' ? `${columnDef.width}px` : columnDef.width,
        initialWidth: typeof columnDef.width === 'number' ? `${columnDef.width}px` : columnDef.width,
        additionalWidth: 0,
        id: index,
      };
```

justrhysism:
I applied this patch manually and it resolved the issue from what I could see.

Issue #2414, opened 2020-09-03
Change visibility of hidden columns breaks the table


urbpeti:
**Describe the bug**
The table goes to error after changing columns visibility. It happens if there is a column with hidden: true and hiddenByColumnsButton: true.
It worked in the previous release.

**To Reproduce**
Serve the Demo app:
1. Add `columnsButton: true` to the table option
2. Add `hidden: true` and `hiddenByColumnsButton: true` at least one of the column definition.
3. Select Column Button and try to set visible the hidden column 
4. Table breaks

You can try out here:
  - #### [StackBlitz](https://codesandbox.io/s/material-table-starter-template-forked-6k16v)

**Expected behavior**
The hidden column should appear.

**Screenshots**
![image](https://user-images.githubusercontent.com/6637121/92156869-e1f1a000-ee29-11ea-8a4b-8e1f9916474c.png)
![image](https://user-images.githubusercontent.com/6637121/92156895-ec139e80-ee29-11ea-917f-4b8e68e8b1c3.png)
![image](https://user-images.githubusercontent.com/6637121/92156957-03528c00-ee2a-11ea-83f8-09785a3f348b.png)

sta2m:
I confirm. Same problem here....

AlexandyZ:
Got the same problem. Hope this can be fixed soon.

Heldeenn:
Got it too. You can install a previous version: 1.64 works just fine !

a-tonchev:
Same problem here

appukuttan-shailesh:
This problem still exists as reported [here](https://github.com/mbrn/material-table/pull/2045) in recent posts.

priyankak10:
Do we have a fix for this. Still breaking the table if trying to show hidden columns

appukuttan-shailesh:
@priyankak10 : might be worth considering shifting to a fork of this project that has resolved this issue:
https://github.com/material-table-core/core

Md-Mudassir47:
follow this [https://codesandbox.io/s/material-table-with-hidden-columns-cndj4?file=/package.json:250-256](https://codesandbox.io/s/material-table-with-hidden-columns-cndj4?file=/package.json:250-256)

i downgraded the version to `1.57.2` and its working fine now

Shahin-ej:
Not working for me evwn i downgraded the version

PR #2468, opened 2020-09-17
Fix for hiddenByColumnsButton property

yann-combarnous:
## Related Issue

#2414 

## Description

Column width was not calculated for columns hidden. This included columns with hiddenByColumnsButton, which can be made visible by user using columnsButton. As a result of missing width, table was crashing.

Approach to resolution is to reset all table columns' width to "undefined" and call setColumns to recalculate columns width when "onChangeColumnHidden" is called.

## Related PRs

N/A

* data-manager


yann-combarnous:
Wrong approach

PR #2480, opened 2020-09-21
disableInternalSorting flag for column and options

drdstewart:
## Related Issue

N/A

## Description

The internal sorting conflicts with server pagination/sorting. Rendered items or computed values can loose position as a result of the `sortList` function. Current work around is to return nothing from `customSort` though this is not ideal. Request is to add `disableInternalSorting` flag to both `columns` and `options` that will still allow for column ordering, but not alter the data. 

## Related PRs

N/A

Sorting


AbbalYouness:
Up

P202120:
So is this working?

Issue #2483, opened 2020-09-22
Hidden actions based on rowData attributes


GabriMagic:
The problem I found is to hide an action based on row attribute values, for instance:

I have this actual feature:

```const actions= [
{
  icon: () => { ... },
  onclick: (event: any, rowData: ? | ?[]) => {...},
  hidden: true
]
```

But if I would like to hide an action based on the table row data I couldn't, this would be my expected feature:

```const actions= [
{
  icon: () => { ... },
  onclick: (event: any, rowData: ? | ?[]) => {...},
  hidden: (rowData) => rowData.disable
]
```
As I saw in the documentation, the definition of the "hidden" attribute is a simple boolean, so I can't use the rowData to change the "hidden" property

![image](https://user-images.githubusercontent.com/6409499/93879848-65d5d400-fcd4-11ea-9b77-954675304ad3.png)

potchy:
I came across the same issue today.
Here's workaround: https://github.com/mbrn/material-table/issues/609#issuecomment-525228694

GabriMagic:
Thanks for your answer, I'm trying to use this with Typescript and I need to type all variables, in this case I sholud declare all as "any" that wouldn't compile:

![image](https://user-images.githubusercontent.com/6409499/93939533-d4418300-fd22-11ea-83e4-d2d2d2699750.png)

Using this workaround woulnd't fix my problem, also when I declare without type, and trying to use it I get lot of errors:

This compile:

![image](https://user-images.githubusercontent.com/6409499/93939873-6ea1c680-fd23-11ea-907c-badae4f78c2e.png)

But not working....

![image](https://user-images.githubusercontent.com/6409499/93939828-592c9c80-fd23-11ea-95de-82d28bc80d4b.png)


So in conclusion, this workaround could fail in some many cases, and also is not compilant with Typescript where you should type all variables, I don't know if I got some error in my code testing it, but it is not working for me.

By the way thanks a lot  for your answer.

GabriMagic:
I did the feature on my own :laughing:, I hope @mbrn could approve it :pray::

https://github.com/mbrn/material-table/pull/2485

zubinshi:
Material table already has build in feature, you can achieve this with following code:

 actions={[
          rowData => ({
            icon:  'toggle_on',
            hidden: rowData.disable
          })
        ]}

GabriMagic:
[...]

Issue #609, opened 2019-05-22
Conditional Actions - styling (hidden?)


natalia-migas:
Hi, 
I would like to completely hide the action button, not only disable it (conditionally). 

![image](https://user-images.githubusercontent.com/25534910/58175341-38bc8000-7ca0-11e9-8199-686b4ef6785e.png)

In docs (https://material-table.com/#/docs/features/actions) is written 

> "You can set an action to disabled or change its color according to rowData", 

but there is no changing color example, so I don't know how could I do that. Maybe this could resolve my problem, by change styling to "display: none".

mbrn:
Hi @natalia-migas ,

You can use display:none css but it is not a effective solution. I may apply a better solution something like disabled(hidden)

mbrn:
This is done with version 1.38.0

buszmen201:
Hey. Thanks for the great module. But I think conditional actions are still missing (so I don't know why it is closed). I found a workaround (since `disabled` prop of Action accepts only boolean (possibly types [TS] are not updated?). Maybe interface Action can accept any function which returns boolean? 
```typescript
export interface Action<RowData extends object> {
  disabled?: boolean | ((data: RowData | RowData[]) => boolean); // here
  icon: string | (() => React.ReactElement<any>);
  isFreeAction?: boolean;
  tooltip?: string;
  onClick: (event: any, data: RowData | RowData[]) => void;
  iconProps?: IconProps;
  hidden?: boolean;
}
```
But okay, here is my workaround

```jsx
<MaterialTable
  actions={[
    (rowData) => {
       return rowData.propToCheck 
         ? { icon: 'custom-icon', disable: true, onClick: (rowData) => { /* anythink */ } }
         : { icon: 'another-icon', disable: false, onClick: (rowData) => { /* anythink */ } }
    }
  ]}
/>

ifndefdeadmau5:
@buszmen201 You saved my ass! thanks

sanjeeviraj:
> Hey. Thanks for the great module. But I think conditional actions are still missing (so I don't know why it is closed). I found a workaround (since `disabled` prop of Action accepts only boolean (possibly types [TS] are not updated?). Maybe interface Action can accept any function which returns boolean?
> 
> ```ts
> export interface Action<RowData extends object> {
>   disabled?: boolean | ((data: RowData | RowData[]) => boolean); // here
>   icon: string | (() => React.ReactElement<any>);
>   isFreeAction?: boolean;
[...]

PR #2485, opened 2020-09-22
[Feature] Add feature to hide actions based on a function with rowDat…

GabriMagic:
…a attributes

## Related Issue

#2483

## Description

Add feature to hide actions with a function in adittion to a simple boolean.

## Related PRs

N/A

- MTableAction

## Additional Notes

I would like to help more this library so I liked to add this feature, it could be added also with "disabled" property. Btw, I wish you could accept it and release the functionality so I could update to the last version.

Also I don't know why the "Prettier CLI" failed, if you could explain that to it would be nice, thanks

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -487,7 +487,13 @@ class App extends Component {
                       icon: "save",
                       tooltip: "Save",
                       onClick: (E, rowData) => alert(rowData.name),
-                      hidden: (rowData) => rowData.name !== "A1",
+                      hidden: (rowData) => {
+                        if (rowData instanceof Array) {
+                          return !!rowData.find((data) => data.name === "A2");
+                        } else {
+                          return rowData.name === "A2";
+                        }
+                      },
                     },
                     {
                       icon: "home",
@@ -529,7 +535,7 @@ class App extends Component {
                     headerSelectionProps: {
                       color: "primary",
                     },
-                    selection: false,
+                    selection: true,
                     selectionProps: (rowData) => {
                       rowData.tableData.disabled = rowData.name === "A1";
 
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -482,6 +482,26 @@ class App extends Component {
               <Grid item xs={12}>
                 {this.state.selectedRows && this.state.selectedRows.length}
                 <MaterialTable
+                  actions={[
+                    {
+                      icon: "save",
+                      tooltip: "Save",
+                      onClick: (E, rowData) => alert(rowData.name),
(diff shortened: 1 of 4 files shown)


GabriMagic:
Morning, I would like to know if I should change something to the pipeline pass successfully, thanks in advance

KannarFr:
@GabriMagic can you check this is working if the selection option is true? It seems function actions are not run if the selection option is set to true.

KannarFr:
@GabriMagic can you confirm that this is not working is the selection option is set to true?

GabriMagic:
"It works", and in this case I think the behaviour is on the hand of the user. For instance...

Having this table and this actions
```actions={[
  {
    icon: "save",
    tooltip: "Save",
    onClick: (E, rowData) => alert(rowData.name),
    hidden: (rowData) => rowData.name === "A2"
  },
  {
    icon: "home",
    tooltip: "Home",
  onClick: (E, rowData) => alert(rowData.name),
    hidden: false,
  },
  {
    icon: "remove",
    tooltip: "Remove",
    onClick: (E, rowData) => alert(rowData.name),
    hidden: true,
  },
]}
```

image

The "save" option won't be available when the name is "A2" so in this case works normally.

---

When we add the "selection" option:
image

You have all actions because the rowData in this case is an Array, it can be easily modified with this action. As I think, the hidden condition for actions are in the hand of the user, if you use the "selection" mode, you should know the rowData will be an array, and you could disable an action when meet a certain condition, or just evaluate the values on the "onChange" or "onSelect".

So I think the feature only add the option to manage the "rowData" inside the **hidden()** mnethod.

```
{
  icon: "save",
  tooltip: "Save",
  onClick: (E, rowData) => alert(rowData.name),
  hidden: (rowData) => {
    if (rowData instanceof Array) {
      return !!rowData.find(data => data.name === "A2");
    } else {
      return rowData.name === "A2";

    }
  },
},
```

KannarFr:
@GabriMagic I fixed a typo on my previous message, can you try to enable the selection: true option which is not available on your example (of answer I received) cf. https://github.com/mbrn/material-table/issues/609#issuecomment-716973634.

GabriMagic:
I add the example with the selection option as "true", you can clone the branch and test it on your own.

WildOsk:
great contribution @GabriMagic I hope it can be implemented ASAP @mbrn

GabriMagic:
Thanks @WildOsk 😄

GabriMagic:
Will this feature reviewed?

PR #2508, opened 2020-10-01
Add support for a summary row

### LINKED ISSUE #27: Add summary row feature
Add summary row feature

### LINKED ISSUE #460: What is the proper way to have a sum row?
Hi-

Could anyone show me what is the proper way to have a sum/subtotal row that can be excluded from column sorting? 
For example:
```
columns={[
                { title: 'Name', field: 'name', defaultSort: 'asc' },
                { title: 'Candy', field: 'candy', type: 'numeric' },
              ]}
data={[{ name: 'Mehmet', candy: 2 }, { name: 'Zerya', candy: 3 }]}
```
Now I want to add a sum row at the end which has "Total       5", and this row should be fixed at the bottom. 
Is there a way to do this?

Thanks in advance!

### LINKED ISSUE #1419: hello guys I need summary row
summary_row
how to like total row 
I cant do it 
Help me Thanks

### LINKED ISSUE #2395: On Filter, show aggregate / total for a column
What I'm trying to accomplish:
- When a user filters the data in my dataset, I would like to show a total for the elements that have been filtered. 

Do you have any recommendations for how you would solve this problem? 

So far, I've tried to add a "total" section into the MLTablePagination component and update the total inside by overriding the "filteredComponent" on the column that's being filtered. So, that when the filter is executed, the total is updated. 

[Example attempt]https://codesandbox.io/s/material-table-starter-template-forked-4tfkn?file=/src/index.js
In the example, the intended behavior is the user starts to type in "Vanilla" and the total updates to the sum of all the records that have the "Vanilla" flavor. The actual behavior for my solution is the user starts to type but the component re-renders (because total is part of state) and causes the filter to be cleared on the re-render

Screen Shot 2020-08-30 at 10 58 22 AM

Any ideas?

Domino987:
## Related Issue

Fixes #1419, Fixes #2395, Fixes #27, Fixes #460

## Description

This PR adds the possibility to add a summary row by adding renderSummaryRow to the props:
image

```
 renderSummaryRow={({ data, index, columns }) => {
                    switch (index) {
                      case 0:
                        return `Total times used: ${data.length}`;
                      case columns.length - 1:
                        return `500`;
                      default:
                        return null;
                    }
                  }}
```
This can be extended to whatever the user needs per column. By skipping certain row indexes, a filter summary can also be achieved.

Additionally, an object with `{value: unknown, style: object}` can be returned to style this cell.
```
return { value: `500`, style: { fontWeight: 800, fontSize: 32 } };
```

\* Table
\* PropTypes
\* Summary-Row
\* TableBody

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -13,6 +13,7 @@ import MTableHeader from "./m-table-header";
 import MTablePagination from "./m-table-pagination";
 import MTableSteppedPagination from "./m-table-stepped-pagination";
 import MTableToolbar from "./m-table-toolbar";
+import { MTableSummaryRow } from "./m-table-summary-row";
 
 export {
   MTableAction,
@@ -30,4 +31,5 @@ export {
   MTablePagination,
   MTableSteppedPagination,
   MTableToolbar,
+  MTableSummaryRow,
 };
--- a/src/components/m-table-body.js
+++ b/src/components/m-table-body.js
@@ -197,13 +197,13 @@ class MTableBody extends React.Component {
       emptyRowCount = this.props.pageSize - renderData.length;
     }
 
+    const columns = this.props.columns.filter((columnDef) => !columnDef.hidden);
+
     return (
       <TableBody>
         {this.props.options.filtering && (
           <this.props.components.FilterRow
-            columns={this.props.columns.filter(
-              (columnDef) => !columnDef.hidden
-            )}
+            columns={columns}
             icons={this.props.icons}
             hasActions={
               this.props.actions.filter(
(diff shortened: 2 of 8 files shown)


phsantiago:
@mbrn @Domino987 guys, anything I can help? We're needing this feature a lot 😞

balanka:
I also need this feature
Can we help to make this PR merge?

TheEcho:
@mbrn Anything we can do to help ?

avala-HFI:
Received a request from a business user to implement something similar on one of our tables.

CaduGimenes:
Guys, I was able to find a way to create a total column using the existing resources in the library:

We need to import "MTableBody" :

`import { MTableBody } from "material-table";
`

In your table component add the attribute _components_

  ```
<MaterialTable
                headers={this.state.header}
                data={this.state.data}
                components={{
                  Body: (props) => (
                    <>
                      <MTableBody {...props} />
                      <tr>
                        <td>TOTAL</td>
                        <td>Your data here</td>
                      </tr>
                    </>
                  ),
                }}
/>
```
In this way, the _component_ attribute is able to override standard library components. So we can overwrite the body of the table and add at the end a row with the data we need, in this case the total...

Hope this helps!
👍

vsn530:
this is not working when we bind remote data instead of static data

PR #2524, opened 2020-10-06
Update Filter types (operator add string)

Chris533:
## Description

I created a dashboard scaffold ```bunadmin``` with ```material-table``` https://github.com/bunred/bunadmin

Custom operators are required for filtering functions, so the type of `operator` needs to be updated.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -73,7 +73,7 @@ export interface MaterialTableProps<RowData extends object> {
 
 export interface Filter<RowData extends object> {
   column: Column<RowData>;
-  operator: "=";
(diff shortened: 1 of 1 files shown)


Domino987 (approved):

nicknijenhuis:
Guess this should not be approved. I do agree that the operator contract is not ideal right now. But this is not the solution.

Chris533:
> This should not be approved. I do agree that the operator contract is not ideal right now. But this is not the solution.

Hello Nick, I agree with you this is not a good solution,  do you have any other suggestions? I really need to customize this operator in many places.

PR #2538, opened 2020-10-12
Add rowData prop to EditField for updating values in real-time

hassaans:
Right now there isn't a way to update any values in real-time using onBlur. If we have rowData available in EditComponent, it can be used to update values in real-time:
```
EditField: props => (
  <TextField
    {...props}
    fullWidth
    onBlur={e => {
      const lines = [...task.lines];
      const lineIndex = lines.findIndex(
        line => line.id === props.rowData.id,
      );
      lines[lineIndex][props.columnDef.field] = e.target.value;
      setTask({ ...task, lines });
      props.onChange(e.target.value);
    }}
    value={props.rowData[props.columnDef.field]}
    inputProps={{ maxLength: 64 }}
  />
),
```

### DIFF
--- a/src/components/m-table-edit-cell.js
+++ b/src/components/m-table-edit-cell.js
@@ -129,6 +129,7 @@ class MTableEditCell extends React.Component {
           <div style={{ flex: 1, marginRight: 4 }}>
             <this.props.components.EditField
               columnDef={this.props.columnDef}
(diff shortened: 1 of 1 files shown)


hassaans:
@mbrn Can you plz look into getting this merged?

PR #2625, opened 2020-11-18
Only validate after touching the field

Domino987:
## Description

This fixes the error that the validation is shown from the beginning.
Should only be validated, after the user interacted with the field.


Delwalt:
Any solution on this issue?

Domino987:
This lib is not maintained anymore. Try the community fork @material-table/core

Issue #2258, opened 2020-07-31
Expanded Multi-level Grouping Collapses When You Select A Row


umeshajb:
**Describe the bug**
I have a table that allows for grouping with multiple columns. I also allow for row selection. The issue I am having is if you group more than one column and select a row within that group the expanded group with the table rows collapses. This is cumbersome for the user if they want to select multiple rows within a group, each time you select a row it will collapse the expanded group and the user will have open the group again to select another row. This seems to be an issue only when you group by multiple columns, it works correctly when you just have one column.

**To Reproduce**

1. Group by multiple columns
2. Expand all the grouping columns
3. Select a row within a group
4. The group expansion collapses and if you need to select another row from the group you will have to expand the group 

**Expected behavior**
When the grouping columns are expanded it should not collapse automatically when you select a row. I would expect it to stay open.

**Desktop (please complete the following information):**

- OS: Mac OS Catalina
- Browser: chrome
- Version 84.0.4147.89

FalkF:
I can confirm this bug, I also might be able to take a closer look at this later.

FalkF:
I took a little dive into `data-manager.js` and could nail the problem down to `findGroupByGroupPath()` returning the wrong group since the `groupsIndex` gets out of sync with  `this.groupedData` in the second dimension of the array. This happens while adding the second grouping somewhere in `sortData()` 

I suspect https://github.com/mbrn/material-table/blob/5e9ef9affbc83db80301c4edee4306aabcc3701f/src/utils/data-manager.js#L958 to be the source of the problem since this is a shallow copy and it would explain why the first grouping is not affected. I could not fix it yet tho, maybe someone with deeper understanding could help here ;)

PR #2630, opened 2020-11-19
Fix expanded multi-level grouping collapsing on row selection

### LINKED ISSUE #2249: Row Selection Within A Multi-Level Group
I have a table that allows for grouping using multiple columns. I also allow for row selection via checkbox. The issue I am having is if you have more than one column for grouping and select a row within that group the expanded group view with the table rows gets collapsed. This is cumbersome for the user if they want to select multiple rows within a group, each time you select a row it will collapse the expanded group and the user will have open the group again to select the row. This seems to be an issue only when you have multiple columns in the grouping, it works correctly when you just have one column.

FalkF:
## Related Issue

#2258 and the already closed #2249 

## Description
The groupsIndex did get messed up while sorting. This leads to the wrong groups being collapsed/expanded on subsequent `getRenderState()` calls. The fix is reindexing the groups after sorting.

- DataManager

## Additional Notes
What a hell of a class :D
Also is the level supposed to start at 1 here?
https://github.com/mbrn/material-table/blob/b7474ca0f5497f04ff3a3d50e709d86de7c639b0/src/utils/data-manager.js#L1000
On a glance 0 seems to make more sense. 

Awesome project 💯

### DIFF
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -983,11 +983,20 @@ export default class DataManager {
 
       this.sortedData = sortGroups(this.sortedData, groups[0]);
 
+      const getGroupsIndex = (groups) =>
+        groups.reduce((result, group) => {
+          result[group.value] = groups.findIndex(
+            (g) => g.value === group.value
+          );
+          return result;
+        }, {});
+
       const sortGroupData = (list, level) => {
         list.forEach((element) => {
           if (element.groups.length > 0) {
             const column = groups[level];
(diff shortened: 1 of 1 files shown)


FalkF:
Could you help me setting up a release branch in my fork so I can use that until this gets eventually merged?
I tried committing the build files but that does not seem to work.
My current workaround is to add my branch to deps, cd into node_modules/material-table, run yarn and yarn build, delete the modules and start.

Issue #2582, opened 2020-11-01
loadingType linear position


Yehudamoskowitz:
Hi, 
When loadingType=linear then the loading div is at the bottom of the table. However, when there are a lot of rows showing and table is reloading then the loading bar is not seen.
Would it be possible to add an option that when loadingType=linear then the loading div should be right under the title before the table rows so that it can be seen even when viewing say 100 rows per page.

Thank you

AravindTReddy:
Any luck on this ? Would be a great asset to have it.

PR #2636, opened 2020-11-21
add a position param for linear loading type

Elabar:
## Related Issue

#2582 

## Description

This PR adds a `linearPosition` field to position the LinearProgress component. Acceptable values are `bottom`(default) and `top`.

This PR should not affect any areas in this application, it simply adds a position control value for linear type loading.

## Additional Notes

This is what it looks like when the `linearPosition` is set to `top`. The loading is also set to absolute position to avoid the table jumping around at the start of the loading and the end of the loading.
image

### DIFF
--- a/src/default-props.js
+++ b/src/default-props.js
@@ -214,6 +214,7 @@ export const defaultProps = {
     headerSelectionProps: {},
     hideFilterIcons: false,
     loadingType: "overlay",
+    linearPosition: "bottom",
     padding: "default",
     searchAutoFocus: false,
     paging: true,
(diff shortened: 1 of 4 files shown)


AravindTReddy:
Does this PR got merged ??

Elabar:
> Does this PR got merged ??

No, and apparently, this library is no longer being maintained actively?
If you really need this feature, you can add a .patch file according to my PR and use [patch-package](https://www.npmjs.com/package/patch-package) to patch your node_modules

PR #2637, opened 2020-11-22
Make bulk edit deleteable

Domino987:
This PR adds a delete action to the edit rows during bulk edit to remove unwanted rows, if onRowDelete is provided.

image

### DIFF
--- a/src/components/m-table-body.js
+++ b/src/components/m-table-body.js
@@ -79,6 +79,8 @@ class MTableBody extends React.Component {
       if (data.tableData.editing || this.props.bulkEditOpen) {
         return (
           <this.props.components.EditRow
+            editable={this.props.editable}
+            actions={this.props.actions}
             columns={this.props.columns.filter((columnDef) => {
               return !columnDef.hidden;
             })}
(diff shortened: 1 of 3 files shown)


mattfrei27:
Any chance of getting this reviewed and merged @mbrn? I could sure use this feature. Thanks!

mattfrei27:
@Domino987 I pulled this commit into my code today and found it doesn't quite meet my need. I have a column in my table that I need to ensure sums to 100. This means that if a user deletes a row that had a nonzero value in that column, the column by definition no longer sums to 100. I'm hoping to solve this issue by using only onBulkUpdate. The thinking being that users can batch together whatever data updates they like (including row deletions) so long as once they are done making updates and click "Save all changes", that column sums to 100. 

Could I please ask you to help me with this? Basically, I need the work you did here to add a delete capability inside a bulk update. But, I don't actually want to trigger any work to be done until the onBulkUpdate fires. This is how the feature to add a row inside bulk update works today. 

Also, I don't want to allow my users to add, modify, or delete a row when they aren't in bulk update mode.

I REALLY appreciate the help here as I'm a bit in over my head trying to make changes to material-table internals myself.

PR #2655, opened 2020-11-28
fix a bug when columns options hidden and hiddenByColumnsButton are set to true

### LINKED ISSUE #2654: columns hidden:true hiddenByColumnsButton:true cause type error
**Describe the bug**
Setting columns hidden and hiddenByColumnsButton to "true" cause an  "type undefined error" when unhiding the column.

Thrown Error:
`Uncaught TypeError: Cannot read property 'match' of undefined`

Where the Error was thrown:
```
var reducePercentsInCalc = function reducePercentsInCalc(calc, fullValue) {
  var captureGroups = calc.match(/(\d*)%/);

  if (captureGroups && captureGroups.length > 1) {
    var percentage = captureGroups[1];
    return calc.replace(/\d*%/, "".concat(fullValue * (percentage / 100), "px"));
  }

  return calc.replace(/\d*%/, "".concat(fullValue, "px"));
};
```
 

**To Reproduce**

1. Create a column
```
columns = {
 { Title: 'some_title',field: 'some_field', hidden:true,hiddenByColumnsButton:true}
}
```
2. Enable buttons to the table
`options = {columnsButton:true}`
3. After rendering the component select the columns button to show the hidden column.

**Expected behavior**
It will show the hidden column

- OS: Windows 10 Pro, 20H2 
- Browser: Chromium Edge, 87.0.664.47 (Official build) (64-bit)
- Version  1.69.2

**Additional context**
What am I am trying to achieve is hide un necessary columns after render to de-clutter the table.
From searching I found some PR but is still pending. Is there any alternative workaround? Please help.

nivranaitsirhc:
## Related Issue
Fixes #2654 

## Description
Fixes a bug when columns options `hidden` and `hiddenByColumnsButton` are set to `true`.

The problem lies in the function `reducePercentsInCalc` when supplied with undefined or null value will throw an error.
I first tried using this approad `calc?.match` but got an error during compiling.
So in the end I choose the primitive approach 
```
if(!calc) return `${fullValue}px`
```

## Related PRs
#2518

All that uses `reducePercentsInCalc` function.
\*

### DIFF
--- a/src/utils/common-values.js
+++ b/src/utils/common-values.js
@@ -10,6 +10,7 @@ export const selectionMaxWidth = (props, maxTreeLevel) =>
   baseIconSize(props) + 9 * maxTreeLevel;
 
 export const reducePercentsInCalc = (calc, fullValue) => {
+  if(!calc) return `${fullValue}px`
   const captureGroups = calc.match(/(\d*)%/);
   if (captureGroups && captureGroups.length > 1) {
     const percentage = captureGroups[1];


kishanio:
Why was this not merged?

nivranaitsirhc:
I don't want to wait for something that will not come

Issue #2519, opened 2020-10-05
Wrong filter value assignment on reordered columns


damianszocik:
**Describe the bug**
Filter values are assigned to the wrong columns after columns has been reordered.

**To Reproduce**
https://codesandbox.io/s/mt-filterreorderingissue-hje43?file=/src/App.js
Drag `Points` column to be the first one and `Qty` to be the second. 
Fill in `Points` column filter.
Typed value appears on the `Surname` column filter input.

**Expected behavior**
Typed filter value should stay in its origin column's filter field.

**Screenshots**
[![mt-columns-reordering.gif](https://s1.gifyu.com/images/mt-columns-reordering.gif)](https://gifyu.com/image/64WX)

**Desktop:**

- OS: Windows 10 Pro 1903
- Chrome 85.0.4183.121 64Bit
- material-table@1.69.1(latest)

Issue #2462, opened 2020-09-16
Filter component breaks after second drag.  


giq-re:
# Guidelines

- #### Please include a demo of the issue/behavior/question you have

- #### Please try to be as detailed as possible

It works after the first drag and then breaks. Meaning the filter stays on the the previous column. I’m not sure where the state is being held for the column order but I know that onorderchange is using the original column ordering as inserted because I had to save that state so so I could ensure that after remote data changes that the sort would be honored.
I will add a demo shortly.  
- #### You may fork one of the following starter templates if you would like:

  - #### [CodeSandbox](https://codesandbox.io/s/material-table-starter-template-xnfpo)
  - #### [StackBlitz](https://stackblitz.com/edit/material-table-starter-template)

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #2658, opened 2020-11-30
Fix: Filter breaks after header is dragged twice

glenkitchen:
## Related Issue

#2462 #2519 

## Description

Fix: After a column is dragged twice, the filter component references the wrong column.

Drag-and-drag of column headers.

## Additional Notes

The current code in the `DataManager.changeByDrag` calculates column order, after sorting the columns by their previous order.   These calculations rely on a mutation (the sort) and are inaccurate after the first reorder.

The new code just swaps the order of the dragged (source) column and dropped (destination) column.
Besides fixing the bug this code is much simpler as it is just a swap algorithm.

### DIFF
--- a/src/utils/data-manager.js
+++ b/src/utils/data-manager.js
@@ -336,21 +336,17 @@ export default class DataManager {
       result.source.droppableId === "headers"
     ) {
       const sourceIndex = result.source.index;
-      const destinationIndex = result.destination.index;
+      const destIndex = result.destination.index;
 
-      if (sourceIndex === destinationIndex) {
+      if (sourceIndex === destIndex) {
         return;
       }
 
-      const sourceColumnOrder = this.columns[destinationIndex].tableData
-        .columnOrder;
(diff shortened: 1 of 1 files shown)


glenkitchen:
Not working. Revisit

madhura-utterwise:
This is still a problem. Do we have a fix for this?

glenkitchen:
I did find a fix for this but I only committed it to a personal fork (which is used in production systems); 
because this project appears dead with no pull requests being accepted.

**Fix**
Change `const sorted = this.columns` to `const sorted = [...this.columns]` in `src/utils/data-manager.js` (line 336).

Used the spread operator to assign a new columns array instead of referencing the existing one i.e. classic mutation bug.  

**Fork**
Here is the [fork](https://github.com/glenkitchen/material-table).
Note that the fork is at v1.68.0 of material table, as this seems to be the last stable version.

Hope this helps!

PR #2692, opened 2020-12-13
Handle custom editComponent for cellEditable tables

### LINKED ISSUE #2559: How to use custom edit component when using cell editable feature ?
I am using Material-Table in my project. I wanted to use a custom edit component when using **cell editable feature**, as I didn't want to use rowUpdate. I tried passing the custom component using "editComponent" prop in columns, but it seems to have no effect.

Here is the codesandbox link of an example implementation - [https://codesandbox.io/s/material-demo-forked-vp145](https://codesandbox.io/s/material-demo-forked-vp145)

If this cannot be done, please consider this as a feature request.

klassm:
## Related Issue

Resolves #2559 - using a custom component for cell editables.

## Description

Before this change setting an editComponent for cellEditable tables had no effect. Now we just render the component a user passes to us.

* cell editables

## Additional Notes

Thank you so much for this project. It really makes life much easier when it comes to displaying table like data. Especially the tree view is so useful. Keep up the great work :-)


klassm:
Sorry for the broken prettier build. Not sure how to fix it though, as it fails already on git clone.

PR #2757, opened 2021-01-25
Add classes styling support for Table, TableHead, TableBody

bernardobgam:
## Avanced Material UI Styling

Added support to define Material UI classes for TableHeader (TableHead and TableRow) and TableBody (TableBody and TableRow) through `headerClasses` and `bodyClasses` in options, providing more styling and customization power.

Example:
```
options={{
    ...otheroptions
    headerClasses: {
        head: {root: classes.headerHead},
        row: {root: classes.headerRow}
    },
    bodyClasses: {
        body: {root: classes.bodyRoot}
        row: {root: classes.bodyRow}   
    }
}}
```

### DIFF
--- a/src/components/m-table-body.js
+++ b/src/components/m-table-body.js
@@ -37,6 +37,7 @@ class MTableBody extends React.Component {
       }
       return (
         <TableRow
+          classes={this.props.bodyClasses.row}
           style={{
             height:
               rowHeight *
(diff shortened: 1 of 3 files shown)


mark-tate:
Or you could change the Material Table components, so that they support a `classes` property, to avoid increasing the size of the API ? 

Anyone wishing to override the styles, can then do so, by wrapping the Material Table components and passing down their own `classes` prop.

PR #2772, opened 2021-02-04
Added MTableEditCell to exported components

alexandreMartinEcl:
## Description

Added a line in types/index so that MTableEditCell can also be imported from the package

## Related PRs

None

None

## Additional Notes

Did not test the build, but simply adding the row on the installed package was enough, so I suppose it should be the same with the original type file.


Dubrzr:
Hi,
I have the same problem, can you fix this please ?
Thank you very much!

PR #2773, opened 2021-02-06
Enable access to the edited state of the row within the editable function

cursedmagician:
Easily enable/disable editing of a column based on other related columns values as they change.

## Related Issue

## Description

Currently the editable function only has access to the row data available when starting editing, by allowing access to the state, you can very easily have dynamic enable/disable of editable columns based on a related value like a checkbox changing.

## Related PRs

| branch              | PR       |
| ------------------- | -------- |
| other_pr_production | [link]() |
| other_pr_master     | [link]() |

\*

## Additional Notes

The change makes the state available as an additional parameter to the editable function, this is to ensure it is a non breaking change in case someone relies on the row data being the original values. While I think this may be overkill, it is safer.

Currently to accomplish something similar, you have to track and maintain a state with custom edit components and then use that state in the editable component, this simplifies the whole process.

### DIFF
--- a/dist/components/index.js
+++ b/dist/components/index.js
@@ -1,0 +1,129 @@
+"use strict";
+
+var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault");
+
+Object.defineProperty(exports, "__esModule", {
+  value: true,
+});
+Object.defineProperty(exports, "MTableAction", {
+  enumerable: true,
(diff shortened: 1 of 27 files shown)


cursedmagician:
Only the original commit was valid, I unfortunately committed other testing changes to the git making it a mess, so deleted the pull request and repository.

PR #2787, opened 2021-02-19
allow to render false Boolean values in type group

thalysonalexr:
## Related Issue

Related with #466 

## Description

I have a Boolean column and I want the label "Sim" to true and "Não" to false, but when doing:

```
render: (rowData, type) => {
        if (type === 'row') return rowData.hasPages ? 'Sim' : 'Não'
        return !rowData ? 'Não' : 'Sim'
},
```
I would like to allow Boolean values ​​to have a render in case the value is false. Works well for rendering in the table, but when grouping the value is lost if false.. I don't know if this was on purpose or a bug.

m-table-cell

### DIFF
--- a/src/components/m-table-cell.js
+++ b/src/components/m-table-cell.js
@@ -27,7 +27,7 @@ export default class MTableCell extends React.Component {
     if (this.props.columnDef.render) {
       if (this.props.rowData) {
         return this.props.columnDef.render(this.props.rowData, "row");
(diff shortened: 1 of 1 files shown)


thalysonalexr:
I solved my problem like this:

image

but I would like to use the checkbox instead of the select box of the lookup because of UX. Keeping a different label for Boolean values ​​instead of the standard check icon. Well, for now it solves my problem.

PR #2809, opened 2021-03-06
This solves the problem of action either showing when checkbox is on/off

EdwinCloud101:
## Related Issue
The issue below was incorrectly closed in my understanding, the bug described there is quite a common requirement.
#980

## Description
The intention of this PR is to add another action toolbar position option that always shows the action in the toolbar.

## Related PRs

no other PRs

It has zero impact on existing applications using material-table

the only component change is:
m-table-toolbar.js

\*

## Additional Notes

This change saved my life, I think it will help other people too. I hope it gets approved, thanks.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -509,7 +509,9 @@ class App extends Component {
                     headerSelectionProps: {
                       color: "primary",
                     },
-                    selection: false,
+                    selection: true,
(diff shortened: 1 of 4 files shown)


EdwinCloud101:
can someone enlight me why the build failed? 

I don't do pull request often, thanks

Issue #2790, opened 2021-02-24
User table interactions are cancelled whenever parent element render is called. Table data updates unnecessarily. 


sydturn:
**Describe the bug**
Technical: componentDidUpdate in material-table does compare against "dirty" data when checking prevProps.data against this.props.data. The props.data the user passes to table is edited by the library (where tableData is appended to it) making it so that prevProps.data will never match this.props.data. 

Visual: User interactions are cancelled on state.data change. Table re-renders unnecessarily when parent element is re-rendered with data unchanged. 

**To Reproduce**

1. Create a component with material table data in state which is updated every 1 second to the same thing it was set to before.
2. Import Material Table component and pass it state.data
3. Ensure Material Table has data in it, and some actions are enabled (for example, perhaps delete)
4. Attempt to delete data from Material Table via UI and observe that after 1 second, once state.data is updated, the "are you sure you want to delete this row?" prompt closes.
5. Similarly this occurs when trying to edit, add, or drop down data in the table. All user interactions close on data.state change, whether data has changed or not. 

**Expected behavior**
User's passed data should not be altered, and should be held onto as a reference to check against-- a new data object should be made by the library so it isn't altering the original data allowing the library to use the original data to check against the incoming data.

Material Table should acknowledge that the data is still the same, and thus should not update their state.data to be props.data

**Screenshots**
On line 693 of Material-Table.js, setDataManagerFields is called, who calls (among other things) dataManager.setData() who loops through each row in data and appends tableData and an ID to the props.data passed in by the user. This is used throughout the codebase for managing a plethora of different states including editing and detail panels and the like. 

![image](https://user-images.githubusercontent.com/18585637/109071267-7f36de00-76c1-11eb-9d6a-80ae1a90e3ea.png)

Unfortunately in the component that is initializing the MaterialTable component, the user doesn't have access to the changes that MaterialTable is making to props.data so whenever the parent component is re-rendered, MaterialTable is handed the data the user has on hand which excludes the tableData info that MaterialTable has in its version of props.data.

Shown in the console at the bottom in my screenshot is the data the user has passed in (which excludes the tableData) and on the right I show the definition of prevProps (the table data in material table's local state prior to this update) which is the same as the data the user passed in but with the addition of tableData object. As such line 688 would always equate to "True" forcing 691 through 695 to trigger a re-render of the table, which closes any interaction the UI user is in the middle of doing (since on a re-render, the tableData object is set back to the dataManager's default values). 

[...]

PR #2818, opened 2021-03-10
Fixed premature user interaction cancellation bug #2790

sydturn:
## Related Issue

Related to bug #2790 

## Description

The ultimate goal was to make sure user interactions aren't being cancelled on data update. User interactions should only be cancelled by the user.

## Related PRs

None

\* MaterialTable
\* DataManager setData

## Additional Notes
DataManager: 
- **Observation**: setData was going through the table data and setting tableData on each element without any regard for the tableData that is already in props.
- **Change**: I changed setData to hold on to the tableData, and append it to the updated data if the updated data's id is the same. This makes it so that if tableData for row 3 had an edit or panel open, then the panel or edit stays attached to that row until the user decides otherwise. I also added the ability for the user to have their own IDs instead of using indexes so that if the user has a panel open on row 2, and deletes row 2, row 3 (whose ID now changed to what row 2 used to have) will open its panel. By allowing the user to pass in their own static IDs that don't change, this scenario is avoidable.

MaterialTable: 

- **Observation**: the same data we remove from columns before comparing data (tableData) was on appended to props.data as well, so the data being passed in from the user, whether it had changed or not, was triggering the propsChanged functionality (setDataManagerFields and setState) which was re-rendering the table and closing user interactions. 
- **Change**: I updated the componentDidUpdate function to clean props.data aswell as props.column and changed the cleanColumns function to have a more generic naming scheme since it is now used for data and column. 

- **Observation**: fast-deep-equal appeared to always return false if there was a render method being passed in with the data. 
- **Change**: I'm not sure if it's because the library can't compare functions or what have you, but since I can't imagine why the render function would change once compiled, I thought it safe to remove the render function in cleanProps just for the compare.

- **Observation**: when the data had changed in componentDidUpdate triggering the propsChanged functionality, setDataManagerFields was calling changeRowEditing  with no rowData or mode which was clearing the tableData editing functionality necessary for keeping user interactions (such as add edit and delete) from closing. 
- **Change**: I removed the call to changeRowEditing  on update just because if the user wanted to cancel their edit, delete, or create, then the user would click the cancel button. User interactions should only be cancelled by the user and we have functionality that does that.


ramonhoyo:
Is there any plan to merge this?

PR #2845, opened 2021-04-06
Upgrade jspdf to mitigate  CVE-2021-23353

jose-antony-by:
## Description

CVE-2021-23353 affects the package jspdf before 2.3.1. ReDoS is possible via the addImage function

PDF Export


jose-antony-by:
For anyone following along: switching to https://github.com/material-table-core/core does fix the issue

Issue #934, opened 2019-08-06
Editing TextField doesn't respect Column Width


ryan-saffer:
**Describe the bug**
I have a column with `type: 'string'`, `editable: 'always'` and the `cellStyle` sets the `width: '60%'`. The column, as well as the text, renders at the correct width when viewing the table. The issue is when I edit a row, the TextField that displays to edit the text has a very small width, and I cannot find a way to set it to match the width of the column.

**To Reproduce**
1. Create a table with an editable string column, set to 60% width.
2. Create a row with a paragraph of text as the content.
3. Click the 'Edit' button.
4. Notice the TextField width shrinks.

**Expected behavior**
The TextField should match the width of the column (and perhaps even wrap as the paragraph gets larger?).

**Desktop (please complete the following information):**
 - OS: Windows 7
 - Browser: Chrome 
 - Version: 74.0.3729.108 [64-bit]

**Additional context**
Not sure if there is a way to fix with custom css..? At this point just looking for a quick fix :)

ryan-saffer:
I have found a solution by overriding the `editComponent` with a material-ui `TextField` where I was able to do my own customization.
```
import TextField from '@material-ui/core/TextField;

...

this.state = {
    columns: [
        {
            ...
            editComponent: props => (
                <TextField
                    value={props.value}
	            fullWidth={true}
	            multiline={true}
	            onChange={e => props.onChange(e.target.value)}
                />
            ...
            )
        }
    ]
}
```
This works fine, but might be nice to be able to provide these `TextField` props (such as `fullWidth` and `multiline`) directly, without overriding the edit component.

jayarjo:
I think `fullWidth` should be simply set to true right away and `multiline` be either a new option for column config or there should be another default type - `text` for example.

@ryan-saffer have you thought about making a Pull Request?

ryan-saffer:
@jayarjo I agree that `fullWidth` should be there by default. I think `multiline` probably better to be an option. Have not made a PR but will look into these options

PR #2854, opened 2021-04-15
multiline and rows props :smile:

Crystal-Moon:
## Multiline is required😢 

TextField has a nice prop - multiline that turns the text field into the auto enlarging textarea.  [#934](https://github.com/mbrn/material-table/issues/934)

## Description

The above issue solution doesn't work with  ["cell editable table"](https://material-table.com/#/docs/features/editable).

I added a `multiline` and` rows` props to the column fixtures which convert input to textarea element naturally. These do not conflict with the 'enterKey' feature defined in [pull#2008](https://github.com/mbrn/material-table/pull/2008)

`multiline` and `rows` properties can be used with all editable tables (input type text) and both are optional 😊 

```javascript
 <MaterialTable
      columns={[
          { title: "Desc1", field: "description", multiline: true, rows: 2 }, // generate a textarea input
      ]}
      data={[
        {
          description: "A long text 2 rows high. a long text 2 rows high. a long text 2 rows high. a long text 2 rows high"
        },
      ]}
      title="Demo Title"
   />
```

### Default values

`multiline: false`     
`rows: 1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,7 @@ class App extends Component {
           columns={[
             { title: "Adı", field: "name" },
             { title: "Soyadı", field: "surname" },
+            { title: "Desc1", field: "description", multiline: true, rows: 2 }, // generate a textarea input
             { title: "Doğum Yılı", field: "birthYear", type: "numeric" },
             {
               title: "Doğum Yeri",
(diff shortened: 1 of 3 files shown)


pouryapb:
this is my solution for multiline text fields:
components={{
          EditField: (props) => (
            <MTableEditField
              {...props}
              multiline={props.columnDef.type === "string" || undefined}
            />
          ),
        }}

PR #2861, opened 2021-04-19
Fixes/remote on search change

orinoco:
## Description

While onChangePage, onOrderChange, onChangeRowsPerPage are all invoked when data is remote, onSearchChange is not, making it difficult to do paging.

Paging, remote data.


phibo23:
This PR should be reopened or resubmitted by an active maintainer because the issue with onSearchChange and remote data still exists

PR #2998, opened 2021-08-28
If e.defaultPrevented, don't trigger onKeyDown behaviour of edit row.

delewis13:
## Related Issue

NA

## Description

Previously there was no way of opting out of the 'Enter', 'Esc' etc. behaviour for the edit row. This was problematic if some inputs in a row expected usage of edit key for example.

By adding a check for defaultPrevented, allow consumers to opt out by adding an `onKeyDown` that calls `e.preventDefault()` within relevant input.


delewis13:
@mbrn WDYT about this PR?

PR #3225, opened 2023-09-28
Update index.d.ts

Pandicool:
## Related Issue

Need helper text variable for getting the warning message in my component

## Description

In my application need to show the warning message in table


SheldonWBM:
If you still need this change, you might be able to request or discuss it at: https://github.com/material-table-core/core
It has more frequent updates but no drastic changes.

PR #3264, opened 2026-01-21
Jspdf critical CVE 2025 68428

prince-inapp:
Fix for material table jspdf critical CVE-2025-68428

### DIFF
--- a/dist/components/index.js
+++ b/dist/components/index.js
@@ -1,0 +1,113 @@
+"use strict";
+
+var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault");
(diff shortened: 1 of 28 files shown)


dist/components/index.js
```diff
@@ -0,0 +1,113 @@
+"use strict";
```
morganjones-avant, line 1: Changes in dist folder will be overwritten with each build.

dist/components/m-table-actions.js
```diff
@@ -0,0 +1,119 @@
+"use strict";
```
morganjones-avant, line 1: Changes in dist folder will be overwritten with each build.

